Issue #16, opened 2015-04-06
Remove old employees


tborres:
Remove Zack from the images

PR #17, opened 2015-04-06
fix(team): Remove Zack

tborres:
Closes: #16

### DIFF
--- a/index.html
+++ b/index.html
@@ -46,49 +46,49 @@ title: Base Two
 </section>
 
 <section id="services" class="container">
-    <div class="section-header">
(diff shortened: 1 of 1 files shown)


tborres:
@mewdriller I removed the photo and info of Zack from the index page. It was a minor change, but the format of the page makes the change look a lot bigger. Let me know if this is good to merge and deploy.

mewdriller:
Looks good to :ship:, unless we want to add Jon and Nick in this as well.

tborres:
I'd add them, but we don't have images yet and I don't know what it should look like without images.

PR #19, opened 2015-04-24
feat(team): add Jon and Nick to the team section

mewdriller:
#### Reviewer Notes

* @tborres: I've added Jon and Nick with low-fi headshots for now. We can replace those once we have a chance to schedule a shoot.

#### Changes In This PR

* Added Jon and Nick to the team section.


mewdriller:
In case it's a pain to run locally for whatever reason, here's what it looks like: 
snapshot

tborres:
:shipit:

PR #95, opened 2017-03-13
feature(base2.io): add jobs and apprenticeship recruiting page

chriswhawkins:


_layouts/jobs.html
```diff
@@ -0,0 +1,47 @@
+<!DOCTYPE html>
```
tborres: Are we planning to provide navigation from the main site to this jobs page?

PR #108, opened 2017-08-04
Adding some initial components to play with

chriswhawkins:
Working through a couple of concepts and getting some components in place to allow people to play around a bit. I would appreciate feedback of any kind, but the design of the actual page is essentially a sandbox for trying out some styling methods.

I'm continuing to work in this same branch and will have updates I plan on pushing over the next few days.


components/Button.js
```diff
@@ -0,0 +1,52 @@
...
+  font-family: var(--font, 'Roboto');
+  font-weight: 400;
+  cursor: pointer;
+  outline: none;
+  padding: .5em 1.25em;
+  min-width: 120px;
+  text-decoration: none;
+  transition: 0.2s ease-in-out;
+
+    &:hover {
```
mewdriller: 🎨: The indentation looks a little off for this `&:hover` (and the `&:active` below) block.

components/Colorfill.js
```diff
@@ -0,0 +1,21 @@
+import styled, { css } from 'styled-components';
+
+const Colorfill = css`
+  &::after {
+    ${props => props.gradient == 'dusk' ? '--color-gradient1: #2C3E50; --color-gradient2: #FD746C;' : null};
+    ${props => props.gradient == 'firewatch' ? '--color-gradient1: #cb2d3e; --color-gradient2: #ef473a;' : null};
+    ${props => props.gradient == 'haikus' ? '--color-gradient1: #fd746c; --color-gradient2: #f44336;' : null};
+    ${props => props.gradient == 'sulphur' ? '--color-gradient1: #CAC531; --color-gradient2: #F3F9A7;' : null};
```
mewdriller: 🎨: This stack of `props.gradient` lines here would be a good opportunity for a use of `_.cond` or something similar.

```
${_.cond([
  [matches({ gradient: 'dusk' }), '--color-gradient1: #2C3E50; --color-gradient2: #FD746C;'],
  [matches({ gradient: 'firewatch' }), '--color-gradient1: #cb2d3e; --color-gradient2: #ef473a;'],
  // ...
])}
```

components/Colorfill.js
```diff
@@ -0,0 +1,21 @@
+import styled, { css } from 'styled-components';
+
+const Colorfill = css`
+  &::after {
+    ${props => props.gradient == 'dusk' ? '--color-gradient1: #2C3E50; --color-gradient2: #FD746C;' : null};
```
mewdriller: 📋: It's not important in this specific usage, but as a general rule we should always be using `===` rather than `==` for comparisons.

components/EmojiButton.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
```
mewdriller: 📋: We should turn on `prettier` / `eslint` for this in a separate task to take care of codestyle standardization.

components/EmojiButton.js
```diff
@@ -0,0 +1,15 @@
...
+import Emoji from './Emoji';
+import Button from './Button';
+
+const EmojiButton = (props) => {
+  return (
+    <Button onClick={props.onClick}>
+      <Emoji styles={{left: 4, margin: 0, height: 20, width: 20, top: -2}}>
```
mewdriller: 🤓: In this case you can safely create these one time outside the component (since it doesn't change based on props):

```
const emojiStyles= { left: 4, margin: 0, height: 20, width: 20, top: -2 };

const EmojiButton = props => (
  <Button onClick={props.onClick}>
    <Emoji styles={emojiStyles}>{props.children}</Emoji>
  </Button>
);
```

.babelrc
```diff
@@ -0,0 +1,8 @@
+{
```
mewdriller: 📋: In some future set of issues we should look into improving the perf of this baseline; for reference here's the Lighthouse report for this branch:

blob-chrome-extension---blipmdconlkpinefehnmjammfjpmpbjk-4e4039ab-cc4f-494c-b3db-e4d60ea8346a

components/Emoji.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Emojify from 'react-emojione';
+
+const Emoji = props =>
+  <Emojify style={props.styles}>
```
kalongthewires: Curious as to how are you determining whether to use inline styles v. styled-component wrappers?

components/EmojiButton.js
```diff
@@ -0,0 +1,28 @@
...
+    left: 4,
+    margin: 0,
+    top: -2,
+    width: 20,
+  },
+};
+
+EmojiButton.propTypes = {
+  children: PropTypes.node.isRequired,
+  styles: PropTypes.objectOf(PropTypes.number),
```
kalongthewires: I didn't even know you could do `objectOf`!

components/Emoji.js
chriswhawkins: Thanks for calling this out. The Emojify component (`react-emojione`) has some default inline styles that I'm basically just overwriting using their same method. Other than here, I'm trying to use styled-component wrappers. It's probably worth a follow up task to refactor the emoji stuff and look into this in the process.

PR #202, opened 2017-12-28
chore(browser-matrix): Support target browsers

### LINKED ISSUE #196: Support target browsers
- [x] Update the `babel-env` settings to make sure we're only transpiling the JS features that aren't supported by our target browsers.
- [x] Ensure that the `styled-components` prefixing (via `stylis`) is sufficient to support our target browsers.
- [x] Investigate CSS linting to avoid using features which aren't supported by our target browsers.

csteinert:
- Added browser list to `.babelrc` and `.stylelintrc`
  - Is there a way to write this once and share it between? idk, help appreciated
- Added Stylelint to project
  - Removed offending CSS, commented out rules where necessary 
  - Overrode rules that we can't abide by due to styled-components

Closes #196

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,4 +1,10 @@
 {
-  "presets": ["react", "es2015", "stage-1"],
+  "presets": ["react", "es2015", "stage-1",
+    ["env", {
+      "targets": {
+        "browsers": ["ie >= 11", "edge >= 16", "last 2 Chrome versions", "safari >= 10.1", "firefox >= 52", "ChromeAndroid >= 62", "iOS >= 10"]
+      }
+    }]
+  ],
   "plugins": ["add-module-exports"]
 }
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,0 +1,20 @@
+{
+  "processors": ["stylelint-processor-styled-components"],
+  "extends": [
+    "stylelint-config-standard",
+    "stylelint-config-styled-components"
+  ],
+  "plugins": [
+    "stylelint-no-unsupported-browser-features"
+  ],
+  "rules": {
+    "declaration-colon-newline-after": null,
+    "value-list-max-empty-lines": null,
+    "declaration-empty-line-before": null,
+    "plugin/no-unsupported-browser-features": [ true, {
+      "browsers": ["ie >= 11", "edge >= 16", "last 2 Chrome versions", "safari >= 10.1", "firefox >= 52", "ChromeAndroid >= 62", "iOS >= 10"],
+      "severity": "warning"
+    }]
+  },
+  "syntax": "scss"
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -1665,7 +1665,7 @@
         "babel-plugin-transform-es2015-unicode-regex": "6.24.1",
         "babel-plugin-transform-exponentiation-operator": "6.24.1",
         "babel-plugin-transform-regenerator": "6.26.0",
-        "browserslist": "2.9.0",
+        "browserslist": "2.10.0",
         "invariant": "2.2.2",
         "semver": "5.4.1"
       }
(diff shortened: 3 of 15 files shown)


.babelrc
```diff
@@ -1,4 +1,10 @@
 {
-  "presets": ["react", "es2015", "stage-1"],
+  "presets": ["react", "es2015", "stage-1",
+    ["env", {
+      "targets": {
+        "browsers": ["ie > 11", "edge >= 16", "chrome >= 63", "safari >= 10.1", "firefox >= 52", "ChromeAndroid >= 62", "iOS >= 10"]
```
MLKendall: I think this should be `ie >= 11`

.stylelintrc
```diff
@@ -4,10 +4,16 @@
...
+  "plugins": [
+    "stylelint-no-unsupported-browser-features"
+  ],
   "rules": {
     "declaration-colon-newline-after": null,
     "value-list-max-empty-lines": null,
-    "declaration-empty-line-before": null
+    "declaration-empty-line-before": null,
+    "plugin/no-unsupported-browser-features": [ true, {
+      "browsers": ["ie > 11", "edge >= 16", "chrome >= 63", "safari >= 10.1", "firefox >= 52", "ChromeAndroid >= 62", "iOS >= 10"]
```
MLKendall: I think this should also be `ie >= 11`.  Otherwise, this will not transpile anything for ie11 (which we are supporting)

MLKendall: Since the latest version of the browser will change as new ones are released, and our browser matrix says that we are supporting "Chrome: latest" would it makes sense to change `chrome >= 63` to `last 2 Chrome versions`?  Perhaps this applies to other browsers in the matrix as well.  @mewdriller Second opinion?

.babelrc
MLKendall: Same as my other comment:

Since the latest version of the browser will change as new ones are released, and our browser matrix says that we are supporting "Chrome: latest" would it makes sense to change chrome >= 63 to last 2 Chrome versions? Perhaps this applies to other browsers in the matrix as well. @mewdriller Second opinion?

.babelrc
```diff
@@ -1,4 +1,10 @@
 {
-  "presets": ["react", "es2015", "stage-1"],
+  "presets": ["react", "es2015", "stage-1",
+    ["env", {
+      "targets": {
+        "browsers": ["ie >= 11", "edge >= 16", "last 2 Chrome versions", "safari >= 10.1", "firefox >= 52", "ChromeAndroid >= 62", "iOS >= 10"]
```
mewdriller: It looks like maybe [this thing](https://github.com/ai/browserslist) will let us share the browser list between the various tools that need it.

.stylelintrc
```diff
@@ -0,0 +1,20 @@
...
+    "declaration-colon-newline-after": null,
+    "value-list-max-empty-lines": null,
+    "declaration-empty-line-before": null,
+    "plugin/no-unsupported-browser-features": [ true, {
+      "browsers": ["ie >= 11", "edge >= 16", "last 2 Chrome versions", "safari >= 10.1", "firefox >= 52", "ChromeAndroid >= 62", "iOS >= 10"],
+      "severity": "warning"
+    }]
+  },
+  "syntax": "scss"
+}
```
mewdriller: It looks like we have some warnings still in here after your work:

```
src/components/atoms/Button.jsx
 14:3  ⚠  Unexpected browser feature "outline" is only partially supported by IE 11   plugin/no-unsupported-browser-features

src/components/atoms/form/Input.jsx
 11:3  ⚠  Unexpected browser feature "outline" is only partially supported by IE 11   plugin/no-unsupported-browser-features

src/components/atoms/form/Textarea.jsx
 4:3  ⚠  Unexpected browser feature "css-resize" is not supported by IE 11, Edge 16, iOS Safari 10.0-10.2,10.3,11.0-11.2   plugin/no-unsupported-browser-features

src/components/atoms/Main.jsx
 6:3  ⚠  Unexpected browser feature "multicolumn" is only partially supported by Firefox 52,53,54,55,56,57   plugin/no-unsupported-browser-features
 7:3  ⚠  Unexpected browser feature "viewport-units" is only partially supported by IE 11                    plugin/no-unsupported-browser-features

src/components/molecules/DescriptiveList.jsx
  9:3  ⚠  Unexpected browser feature "flexbox" is only partially supported by IE 11                           plugin/no-unsupported-browser-features
 44:3  ⚠  Unexpected browser feature "multicolumn" is only partially supported by Firefox 52,53,54,55,56,57   plugin/no-unsupported-browser-features
 46:3  ⚠  Unexpected browser feature "multicolumn" is only partially supported by Firefox 52,53,54,55,56,57   plugin/no-unsupported-browser-features
```

Is our policy that warnings are OK to merge in but not errors? Some of these seem a bit concerning, but I'm not sure what your thoughts on it are.

Issue #197, opened 2017-12-22
Form and Contact Information Layout


MLKendall:
Dependencies: #185, #181

Display the form and contact information per the design documents
- [x] Layout should display correctly on desktop, tablet and mobile
- [x] Should navigate into view when the "Make Contact" link is clicked.
~* Purchase the design assets listed in the document~ (Moved to #204)

https://www.figma.com/file/YApYHX8kQ0WI0i7NcxnrY2vL/BaseTwo-Contact-Page

![image](https://user-images.githubusercontent.com/8663544/34311653-c52eb022-e72c-11e7-8994-2ed5f2cf50df.png)

PR #203, opened 2018-01-02
feat(contact-us): Contact Us header and background

### LINKED ISSUE #185: Contact page header and background
* [x] Styles and markup for the Contact Us page header.
* [x] Display background image for Contact Us.
* [x] Implement Make Contact button.
* [x] Should work on mobile and desktop.
~* [ ] Purchase design assets listed in the design below.~

image

Design: https://www.figma.com/file/YApYHX8kQ0WI0i7NcxnrY2vL/BaseTwo-Contact-Page

csteinert:
Items of note:

- I was too ambitious with Grid, previously. We don't need it, this branch removes it. It will be missed. ⚱️ 
  - Doing so caused some rework in the Hero section.
- `gradient` is now a styling option for Button.

Closes #185


src/components/atoms/BaseTwoLogo.jsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import styled, { css } from 'styled-components';
+
+const circleMixin = css`
+  background: black;
```
kalongthewires: Should this be #000 for consistency?

src/components/atoms/BaseTwoLogo.jsx
```diff
@@ -0,0 +1,39 @@
...
+  display: flex;
+  flex-direction: row;
+`;
+
+const BaseTwoLogo = () => (
+  <LogoWrapper>
+    <SolidCircle />
+    <StripedCircle />
+  </LogoWrapper>
+);
```
kalongthewires: Fancy.

src/components/atoms/Button.jsx
```diff
@@ -1,26 +1,40 @@
...
-  background-color: ${themed('color.primary')};
-  border: 2px solid #80cbc4;
-  border-radius: 2px;
-  color: ${flow(themed('color.primary'), readableColor)};
+  ${themed('font.sansSerif')};
+  background-color: ${themed('color.transparent')};
+  border: 1px solid ${themed('color.link')};
+  color: ${themed('color.link')};
   cursor: pointer;
+  font-size: 2.625em;
```
kalongthewires: Button is disproportionately large of mobile. Looks like text elsewhere is in rems.

src/components/atoms/Button.jsx
```diff
@@ -38,7 +52,7 @@ const Button = styled.button`
 
   &:focus,
   &:hover {
-    background-color: ${flow(themed('color.primary'), lighten(0.1))};
+    background-color: ${themed('color.linkHover')};
```
kalongthewires: Wondering if the color contrast between the background and the text is sufficient? Looks like the submit button in the form has a different hover effect as well -- are they meant to be different?

src/components/organisms/ContactHeader.jsx
```diff
@@ -0,0 +1,134 @@
...
+        </LocationText>
+        <Earth src="img/planet-earth.jpg" alt="Planet Earth" />
+        <LocationTagline>
+          Thanks to satellites and the internet, we can work with you wherever
+          you are.
+        </LocationTagline>
+        <Button gradient onClick={onContact}>
+          Make Contact
+        </Button>
+        <Scroll src="img/scroll-arrows.svg" alt="" />
```
kalongthewires: The scroll arrows under the satellite in the hero are small compared to the arrows here (particularly noticeable on mobile). Not sure if that's intentional.

Also, should this have alt text?

src/components/organisms/ContactHeader.jsx
```diff
@@ -0,0 +1,134 @@
...
+        <LocationText>
+          is located in <strong>Columbus, OH</strong> and{' '}
+          <strong>Pittsburgh, PA</strong>.
+        </LocationText>
+        <Earth src="img/planet-earth.jpg" alt="Planet Earth" />
+        <LocationTagline>
+          Thanks to satellites and the internet, we can work with you wherever
+          you are.
+        </LocationTagline>
+        <Button gradient onClick={onContact}>
```
kalongthewires: Should clicking the button do something? Or is this functionality being added separately?

src/components/organisms/ContactHeader.jsx
```diff
@@ -0,0 +1,134 @@
...
+    margin: 0 auto;
+    max-width: 90%;
+    min-width: 10em;
+
+    ${mediaQuery.small`
+    max-width: 600px;
+  `};
+
+    ${mediaQuery.medium`
+    max-width: 615px;
```
kalongthewires: Seeing a lot of different units for things. Wondering if there's some rule you're going by to determine whether to use px, vws, rem, or em?

src/components/organisms/Hero.jsx
```diff
@@ -3,79 +3,51 @@ import styled from 'styled-components';
...
-  grid-row-end: 2;
+  height: 100vh;
+  padding: 3.75em 0 0 0;
+  position: relative;
+  text-align: center;
 
   ${mediaQuery.small`
     background-image: url('img/backgrounds/hero/b2-hero-background-960.png');
     background-size: 120%;
+    min-height: calc((100vw * ${BGAspectRatio}) * 1.2);
```
kalongthewires: What is the 1.2 value here? Magic number?

src/components/atoms/BaseTwoLogo.jsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
```
kalongthewires: Should B2LogoText be renamed to BaseTwoLogoText for consistency?

src/components/organisms/Hero.jsx
csteinert: Aspect ratio on the background image.

src/components/atoms/Button.jsx
csteinert: Not sure, to my knowledge there wasn't a mock of a hovered button. I set the submit hover styles to be the default, so now all buttons have them. 👍

src/components/organisms/ContactHeader.jsx
csteinert: The PX were lazy, I'm replacing them now with the `em` function from Polished.

My thought process around em, vws, or % is:
- Use EMs when I know the exact size I want an element to be/not to exceed (translate from px if necessary) or when sizing typography relative to the 16px base size.
- Use VW/VH for high-level layout elements where viewport size is a design factor (eg our background images).
- Use percentages to resize elements relative to the parent size.

And now that I've written that, I'll make sure I abide by it! 😂

csteinert: I think the REMs were coming from Mandy's code. I never know which way to go with those. Gonna stick with EMs for now, can discuss when she's back.

src/components/organisms/ContactHeader.jsx
csteinert: It should navigate to the contact form below, probably via a link to an ID (which won't exist until the bottom section is done). But now I'm having an existential crisis where I don't think this should be a button element. It's a link. Might have to make a new element of "link button" or something to make this work properly.

Can I punt this to #197?

kalongthewires: Agree and yes.

src/components/organisms/ContactHeader.jsx
kalongthewires: 😆 Sounds good to me.

Issue #215, opened 2018-01-10
Make Contact Button too Large in IE


MLKendall:
The Make Contact button is too wide in IE

![ie centering issue](https://user-images.githubusercontent.com/8663544/34782712-8bf7a0d2-f5f7-11e7-9521-d40556d3829c.PNG)

For reference, the correct button size can be viewed here:
https://www.figma.com/file/yfECZFxj1aZBSxKk4q523K/BaseTwo-Contact-Page

PR #219, opened 2018-01-10
(fix) 214 Contact Us Text Center IE

### LINKED ISSUE #214: Text in contact section not centered in IE
The line "Thanks to satellites and the internet, we can work with you wherever you are." is not centered in IE.

ie centering issue

For reference, the correct appearance can be viewed here:
https://www.figma.com/file/yfECZFxj1aZBSxKk4q523K/BaseTwo-Contact-Page

MLKendall:
Added widths to LocationTagline and ScollButton

Although dealing with widths, may be related to known IE flexbox bug:
https://connect.microsoft.com/IE/feedback/details/802625/min-height-and-flexbox-flex-direction-column-dont-work-together-in-ie-10-11-preview

Closes #214 and #215

### DIFF
--- a/src/components/organisms/contact/ContactHeader.jsx
+++ b/src/components/organisms/contact/ContactHeader.jsx
@@ -73,8 +73,8 @@ const LocationText = styled.span`
 const LocationTagline = styled.p`
   font-style: italic;
   margin: 0 auto 2em;
-  width: 100%;
   max-width: 17em;
+  width: 100%;
 
   ${mediaQuery.small`
     font-size: 2em;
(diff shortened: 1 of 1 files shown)


src/components/organisms/contact/ContactHeader.jsx
```diff
@@ -73,6 +73,7 @@ const LocationText = styled.span`
 const LocationTagline = styled.p`
   font-style: italic;
   margin: 0 auto 2em;
+  width: 100%;
```
csteinert: Alphabetical order of properties pls (we should be linting for that if we're not, I'll look into it)

csteinert:
Looks like we lost the alignment on the scroll icon: 
screenshot 2018-01-11 11 25 24

src/components/organisms/contact/ContactHeader.jsx
MLKendall: This is actually for the alignment comment (Github won't let me comment there)

On my IE, the alignment is correct.  I asked Kim for a third check, because I don't know if the difference is with me or with browserstack.

ie check

PR #224, opened 2018-01-18
feat(contact-us): Thank You page

### LINKED ISSUE #187: Thank you page
When the user submits the Contact Us page form, a page with a Thank You message should be displayed. Should be available on mobile and desktop.

csteinert:
- Added background and text styles
- Removed team list

Closes #187

### DIFF
--- a/src/pages/thanks.jsx
+++ b/src/pages/thanks.jsx
@@ -8,11 +8,25 @@ import {
   P,
   Section,
 } from '../components';
+import { mediaQuery } from '../util/style';
 
 const Wrapper = styled(Main)`
   background-image: url('img/backgrounds/hero/b2-hero-background-1920.png');
   background-position: 50% 0;
+  background-size: 270%;
   height: 100vh;
+
+  ${mediaQuery.small`
+    background-size: 140%;
+  `};
+
+  ${mediaQuery.medium`
+    background-size: 100%;
+  `};
+
+  ${mediaQuery.large`
+    background-position: 50% 5%;
+  `};
 `;
 
 const ThanksContent = styled(Section)`
@@ -22,6 +36,8 @@ const ThanksContent = styled(Section)`
   flex-direction: column;
   height: 100%;
   justify-content: space-between;
+  margin: 0 auto;
+  max-width: 1024px;
   padding-bottom: 2em;
   text-align: center;
 `;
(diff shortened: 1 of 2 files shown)


src/pages/thanks.jsx
```diff
@@ -1,82 +1,107 @@
...
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+  justify-content: space-between;
+  margin: 0 auto;
+  max-width: 1024px;
+  padding-bottom: 2em;
+  text-align: center;
+`;
+
```
MLKendall: This gets really wide on my screen.  Can we maybe give it a max-width?

src/pages/thanks.jsx
```diff
@@ -1,82 +1,107 @@
...
 } from '../components';
-import { toNodesWithImage } from '../util/graphql';
+import { mediaQuery } from '../util/style';
+
+const Wrapper = styled(Main)`
+  background-image: url('img/backgrounds/hero/b2-hero-background-1920.png');
+  background-position: 50% 0;
+  background-size: 270%;
+  height: 100vh;
+
```
MLKendall: Did you intend for the background to repeat on wider screens?  I am seeing two partial earth on either side on my screen.  I played with adding background-repeat: no-repeat and was not at all offended by having the background fade to black on the sides, so maybe add that?

src/pages/thanks.jsx
csteinert: There is a max-width on the content...? Is it too wide still? Can I get a screenshot?

src/pages/thanks.jsx
```diff
@@ -1,82 +1,107 @@
...
+    font-size: 2em;
+  `};
+
+  ${mediaQuery.medium`
+    font-size: 2.25em;
+  `};
+`;
+
+const GoHome = styled(CallToAction)`
+  flex-basis: auto;
```
MLKendall: On larger screens, I feel like maybe the button should be more towards the top (and perhaps a bit thinner).  I don't like looking down to the bottom of the screen to see the Return Home button.

Not necessarily exactly like this, but maybe something like it?
fireshot capture 26 - base two - http___localhost_8000_thanks

src/pages/thanks.jsx
mewdriller: I just set my window size to 6000x6000 in the Chrome devtools "responsive" device section and wasn't able to see the issue you're talking about. Maybe it's already been resolved by @csteinert?

csteinert: I'm only seeing it repeat vertically on large screens, but I'll set it to no-repeat and then we shouldn't have any problems in either direction.

MLKendall:
Text extends off page in IE
ietext

PR #226, opened 2018-01-29
Blog

### LINKED ISSUE #174: Blog
Blog landing page and individual blog posts. Available on mobile and desktop.

csteinert:
Added themes to site, created light theme for blog.

Closes #174

### DIFF
--- a/src/components/atoms/form/Input.jsx
+++ b/src/components/atoms/form/Input.jsx
@@ -4,7 +4,7 @@ import { cssSome, themed } from '../../../util/style';
 const Input = styled.input`
   background: 0 0;
   border: none;
-  border-bottom: 1px solid ${themed('color.accent')};
+  border-bottom: 1px solid ${themed('color.spaceGreen')};
   color: ${themed('color.text')};
   display: block;
   margin: 0;
--- a/src/components/atoms/form/Input.jsx
+++ b/src/components/atoms/form/Input.jsx
@@ -4,8 +4,8 @@ import { cssSome, themed } from '../../../util/style';
 const Input = styled.input`
   background: 0 0;
   border: none;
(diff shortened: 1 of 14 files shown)


src/components/atoms/form/Input.jsx
```diff
@@ -4,16 +4,16 @@ import { cssSome, themed } from '../../../util/style';
...
   display: block;
   margin: 0;
   padding: 4px 0;
   text-align: left;
   width: 100%;
 
   ${cssSome('isDirty', 'isFocused')`
-    box-shadow: 0 1px 0 0 ${themed('color.spaceGreen')};
+    box-shadow: 0 1px 0 0 ${themed('color.accent')};
   `};
```
MLKendall: Looks like color.accent is spaceBlue and this is supposed to be spaceGreen here.

src/components/molecules/GlobalNavigation.jsx
```diff
@@ -82,20 +82,20 @@ const NavItem = styled(LI)`
 `;
 
 const NavLink = styled(Link)`
-  color: ${themed('color.spaceBlue')};
+  color: ${themed('color.navTextSmall')};
   display: inline-block;
```
MLKendall: Is there a different name we could use here?   When I first saw this, it seemed like it meant that the text was supposed to be small, not the screen size.  I don't know if navTextMobile or something similar would work better?  (Not sure if I like navTextMobile since we might eventually have to also have a collapsible menu on tablet as well.)

csteinert: Excellent point. Maybe collapseNavText?

MLKendall: I think that would work.

PR #233, opened 2018-02-09
feat(services): Services implementation

### LINKED ISSUE #172: Services
Display Services we provide at Base Two. Available on mobile and desktop.

csteinert:
Closes #172


src/pages/index.jsx
```diff
@@ -1,18 +1,76 @@
...
 
 class IndexPage extends React.Component {
   static defaultProps = {};
 
+  static propTypes = {
+    services: PropTypes.arrayOf(
+      PropTypes.shape({
+        id: PropTypes.node.isRequired,
+        image: PropTypes.shape({}).isRequired,
+        imageAlt: PropTypes.string.isRequired,
```
MLKendall: This is resulting in a prop types console error.  Should be imgAlt, not imageAlt.

src/components/molecules/ServiceList.jsx
```diff
@@ -0,0 +1,141 @@
...
+const ServiceDescription = styled(P)`
+  font-size: ${em('18px')};
+  margin: 0;
+
+  ${mediaQuery.small`
+    font-size: ${em('20px')};
+  `};
+
+  ${mediaQuery.medium`
+    font-size: ${em('32px')};
```
MLKendall: My fault in the Figma doc - this should be 28px, not 32px.  I think it looks better with smaller text here.

src/components/molecules/ServiceList.jsx
```diff
@@ -0,0 +1,141 @@
...
+
+    &:nth-of-type(even) {
+      flex-direction: row-reverse;
+    }
+  `};
+`;
+
+const ServiceList = styled.ul`
+  ${containerSizing};
+  background: url('img/backgrounds/StarFieldTransparent.png') repeat;
```
MLKendall: Can this be 100% on the x -axis and repeat on y only?  It ends up having some off empty areas when it repeats x.

There are also places where the stars cut off.  I think that is in the file.  I will fix those and get the new one to you.

stars overlap

MLKendall: Also - there is no purple fog in the background.  I made them separate images, but I can also combine them if that is easier (and maybe saves downloading two assets.

MLKendall: Starfield Background starts too close to the monitor.  Can we put some padding in there to move it down a bit?
starstooclosetomonitor

MLKendall: ☝️ Reduce space between image and text content.

Issue #252, opened 2018-02-28
Services Section Does not Render Properly in Microsoft Edge


MLKendall:
The Services Section is not rendering properly in MS Edge.  The individual items render too narrow.

![services in edge](https://user-images.githubusercontent.com/8663544/36796131-d4b4989c-1c72-11e8-8649-66d683b83492.PNG)

PR #263, opened 2018-02-28
#252 Service section in Edge issue

csteinert:
Updated max-width and added width to service list items to fix issue in Edge

### DIFF
--- a/src/components/molecules/ServiceList.jsx
+++ b/src/components/molecules/ServiceList.jsx
@@ -1,9 +1,13 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import styled from 'styled-components';
-import { em } from 'polished';
+import { em, rem } from 'polished';
 import { H3, Img, P, LI, TwoToneHeading } from '../atoms';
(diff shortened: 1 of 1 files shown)


src/components/molecules/ServiceList.jsx
```diff
@@ -1,9 +1,13 @@
...
 import React from 'react';
 import styled from 'styled-components';
-import { em } from 'polished';
+import { em, rem } from 'polished';
 import { H3, Img, P, LI, TwoToneHeading } from '../atoms';
-import { containerSizing, mediaQuery } from '../../util/style';
+import { containerSizing, mediaQuery, themed } from '../../util/style';
+
+const fadeOut = {
+  value: `${em('150px')}`,
```
MLKendall: LGTM, but what does this do? ☝️

csteinert: It is a linear gradient that goes from transparent to black, placed on an `after` pseudo-element, that fades out the star background to keep it from cutting off mid-star when the section ends.

PR #345, opened 2018-10-03
fix: add propType definitions to RadioButton

tborres:


src/components/atoms/form/RadioButton.jsx
```diff
@@ -15,13 +16,21 @@ const RadioInput = styled.input.attrs({
...
+function RadioButton({ id, label, ...rest }) {
   return (
     <Label htmlFor={id}>
       <RadioInput {...rest} id={id} />
       {label}
     </Label>
   );
 }
+RadioButton.defaultProps = {
+  id: nanoid(),
```
mewdriller: 🤔 I believe this will create one unique ID from `nanoid` that is the default for all instances of `RadioButton` that don't supply their own `id` prop. Were you trying to have it default per-instance? If so, you can use an inline default in the props to get that per-render:

```jsx
function RadioButton({ id = nanoid(), label, ...rest }) {}
```

src/components/atoms/form/RadioButton.jsx
```diff
@@ -15,13 +16,21 @@ const RadioInput = styled.input.attrs({
...
+};
+
+RadioButton.propTypes = {
+  id: PropTypes.string,
```
mewdriller: 📝 There's a convention (not enforced via `eslint`) that the required prop-types are listed first (in alphabetical order), followed by a blank line, followed by the optional prop-types (in alphabetical order).

src/components/atoms/form/RadioButton.jsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import styled from 'styled-components';
 import nanoid from 'nanoid';
+import PropTypes from 'prop-types';
```
mewdriller: 📝 I personally try to keep the imports alphabetized by their package name and/or relative path to avoid as much diff-churn as we can. Unfortunately none of the you-would-think-they-work `eslint` plugins actually enforce that. 

I'm not sure if that's a convention the rest of the team is interested in on this project though, so feel free to ignore this comment.

PR #351, opened 2018-10-03
Create Job Form

jridenbaugh:
Adds component for job form. Works for both apprenticeships and regular jobs. Still might need some styling work. The file upload certainly needs styling work,.


src/components/organisms/jobs/index.js
```diff
@@ -0,0 +1,4 @@
+import JobForm from './JobForm';
```
reeddunkle: We started switching these `import`+`export`s over to:

```js
export { default as JobForm } from './JobForm';
```

PR #353, opened 2018-11-06
feat(jobs): Jobs body content

### LINKED ISSUE #332: Create Jobs Body (Organism)
Create the body of the Jobs page as an organism using atoms and molecules.

jobs_body_organism

* https://www.figma.com/file/KFeaA4xNivQ7t9xvrNrjkQ/Jobs-For-Backlog?node-id=0%3A1

### LINKED ISSUE #350: Refactor Jobs Header
We could refactor the Jobs Header so that we don't have a `JobsHeaderSmall` and `JobsHeader` (which as action buttons).

Sean pointed out:

> We could probably abstract the HeaderActions portion of the JobsHeader out into a JobsHeaderActions and then just have a single JobsHeader that takes children throws them below the SubText. Then we'd be down to just a single JobsHeader.
> There are couple small style changes relating to bottom-margin in JobsHeaderSmall it looks like.

kalongthewires:
* Adjust JobsHeader to match body content

Closes #332
Closes #350


_content/jobs.json
```diff
@@ -0,0 +1,16 @@
...
+    "url": "#"
+  },
+  {
+    "id": "2",
+    "position": "Developer",
+    "description":
+      "Elegant user interface design is an integral part of our work. Using a variety of methods including (but not limited to) user interviews, wireframes, and prototypes, we ensure that the end user’s experience is efficient and enjoyable.",
+    "url": "#"
+  }
+]
```
kalongthewires: These are sample jobs that will need to be removed.

src/components/organisms/jobs/JobsContent.jsx
```diff
@@ -0,0 +1,152 @@
...
+          src={officeCollageImage}
+          alt="Our Base Two team"
+          width="1000"
+        />
+      </JobsSection>
+      <JobsSection>
+        <H2>We believe in balance</H2>
+        <P largeText>
+          We offer great benefits so that the
+          &lt;strong&gt;life&lt;&#47;strong&gt; portion of work-life balance can
```
csteinert: screenshot 2018-11-14 13 40 44

I assume we want this to actually be bold, but if not maybe wrap strong in `code` tags?

csteinert:
I think the absence of `jobs.json` is what's causing the checks to fail. I'm down with leaving the placeholder jobs in there for now, if it's too much extra effort to account for a lack of jobs at this time.

kalongthewires:
Turns out you can't currently have empty data of any sort or Gatsby chokes on it (so an empty json file or a json file with [] both cause errors). They're working on a fix and basically it sounds like people are just keeping dummy data in place in the meantime.

https://github.com/gatsbyjs/gatsby/issues/3344

I'm going to add the dummy data back in and merge.

Issue #355, opened 2018-12-04
Initial load time


tborres:
The initial load of the homepage (specifically the large earth image) takes too long (roughly 5 seconds).

**Dev Notes**
Be sure to do a hard refresh when testing (ctrl+shift+r).
We should serve properly-sized images for the requesting device

Issue #162, opened 2017-12-06
Reduce render-blocking stylesheets


mewdriller:
Our Lighthouse "Performance" report includes a note that we could save ~600ms in our "First meaningful paint" by reducing render-blocking stylesheets.

Specifically, it calls out `/css?family=Roboto:400,500` with a reference to [this help article](https://developers.google.com/web/tools/lighthouse/audits/blocking-resources).

PR #357, opened 2018-12-17
355 initial load time

chriswhawkins:
Closes: #355, #162 

## Steps taken:
1. Optimized all the images I could find (started with those named in the lighthouse report and identified any similar uses)
2. Created `.webp` versions of large images (again, starting with those identified in the audit)
3. Updated `Earth` component in `ContactHeader` to use `picture` because it allows for multiple sources, aka `webp` with fallback to `jpg`.
4. Created a "mixin" to allow for assigning similar fallbacks in css background images, leveraging css @supports to identify `webp` capable browsers.
5. Implemented the `webpBackground` "mixin" in all background image use cases I could find.
6. Combined google font families to a single request url
7. Added async attribute to linked polyfill.io js

## Comments:
 - I think the image optimization is pretty obvious, but I'd like to see this documented or automated.
 - It feels like there has to be a cleaner (gatsby plugin) way of handling srcSet in components (no. 3 above)
 - It feels like there has to be a better way to handle the css background webp support. I know modernizr can add classnames based on browser capabilities, for example, but this felt light and straight forward for a first pass. (no. 4/5 above)

## Lighthouse Audit Before:
image

## Lighthouse Audit After:
image


src/html.jsx
```diff
@@ -86,18 +86,14 @@ class Html extends React.Component {
...
-            rel="stylesheet"
+            href="https://fonts.googleapis.com/css?family=Roboto:100,100i,400,500|Anonymous+Pro"
+            rel="prefetch"
           />
           <title>Base Two</title>
           {headComponents}
           {css}
 
-          <script src="https://cdn.polyfill.io/v2/polyfill.min.js" />
+          <script async src="https://cdn.polyfill.io/v2/polyfill.min.js" />
```
dmiller-nj: By moving this to a `async` script rather than a blocking one, have we opened up the possibility of the polyfill not being received/parsed/evaluated before some browser incompatible JS runs?

dmiller-nj:
Gatsby has some image-related features that might address some of your questions/comments: https://using-gatsby-image.gatsbyjs.org/. Let me know if you want to follow-up on this and we can hop on a Zoom.

src/html.jsx
chriswhawkins: Yeah you're right. It looks like the polyfill.io api can take a callback function that would solve for that concern. I'll look into that and potentially roll back the change here and open a new issue.

src/components/molecules/ServiceList.jsx
```diff
@@ -3,30 +3,36 @@ import React from 'react';
...
-  background: url('img/backgrounds/star-field.png');
   display: flex;
   flex-direction: column;
   list-style: none;
   padding-left: 0;
   padding-top: ${themed('navHeight.value')};
   position: relative;
   text-align: center;
+
+  ${webpBackground('img/backgrounds/star-field.png', 'img/backgrounds/star-field.webp')};
```
csteinert: Total nitpick, BUT, sometimes the sources are on one line vs. separate lines. Can we pick one way and go with it? 🙏

src/components/organisms/contact/ContactHeader.jsx
```diff
@@ -42,7 +42,7 @@ const Wrapper = styled.div`
   }
 `;
 
-const Earth = styled.img`
+const Earth = styled.picture`
```
csteinert: Very exciting, but we are supporting IE11... 😬 

We could perhaps stop supporting IE11? @MLKendall, thoughts?

tborres: I hadn't noticed any issues in IE11. Does this cause an issue?

As for should we support IE11, I think yes. A lot of businesses and business people are stuck on that browser, and there's a good chance that if they come to our site that they will be visiting it with IE11.

chriswhawkins: @csteinert and @MLKendall - I also agree that we should keep requirements around IE11, but I might be missing something here. Are either of you seeing an issue with the earth image?

While the `picture` element may not have support in IE11, the spec for it requires a nested `img` to account for older browsers.

src/components/molecules/ServiceList.jsx
chriswhawkins: It's only wrapped in some places b/c of line length settings, but I can just do it everywhere for consistency 👍

src/components/organisms/contact/ContactHeader.jsx
MLKendall: Everything seems to be working fine on IE on my end.  I see what you are talking about with the nested .  I think this is fine.

src/html.jsx
reeddunkle: From the [docs](https://polyfill.io/v2/docs/examples#feature-detection):

> To make sure the polyfills are present before you try to run your own code, you can attach an `onload` handler to the https://cdn.polyfill.io script tag, use a more sophisticated script loader or simply use our `callback` argument to evaluate a global callback when the polyfills are loaded:

```html
<script src="https://cdn.polyfill.io/v2/polyfill.min.js?callback=polyfillsAreLoaded" defer async></script>
<script>
  function polyfillsAreLoaded() {
    console.log('Now to do the cool stuff...');
  }
</script>
```

chriswhawkins: async removed from this branch/pr and new issue opened to address using the suggested method #371

chriswhawkins:
> Gatsby has some image-related features that might address some of your questions/comments: https://using-gatsby-image.gatsbyjs.org/. Let me know if you want to follow-up on this and we can hop on a Zoom.

New issue created for implementing the gatsby plugin #372

PR #362, opened 2019-01-03
style(blog): New Header for Blog Posts

### LINKED ISSUE #360: Implement Header for Blog Posts -- #358
Implement the header section of the new Blog Posts Design

Acceptance Criteria

Story #358 

- [ ] Works on Desktop, Tablet, Mobile
- [ ] Works in IE, Firefox, Chrome, Safari

Figma Designs Can be Found Here: 
https://www.figma.com/file/SwgRtCNSDlbuSkBZJyMfD6/Blog-Post-Design?node-id=0%3A1

image

MLKendall:
Closes #360


src/util/style.js
```diff
@@ -23,6 +23,7 @@ const mq = (target, ...conditions) => (...args) => css`
 const mediaQuery = {
   xsmall: mq('screen', 'max-width: 480px'),
   small: mq('screen', 'min-width: 481px'),
+  smedium: mq('screen', 'min-width: 767px'), // Needed an intermediate size for global navigation
```
kalongthewires: I would remove the comment because this break point is already being used in post.jsx, not just global navigation.

kalongthewires: I worry that anyone coming into the code base new will think this is a typo. I certainly did until I saw this change in style.js. It also feels like a code smell -- like there is some tech debt or non-flexible ways of styling in the app that are now requiring a sixth breakpoint. 

That said, I think any changes to avoid needing the extra breakpoint would be pretty wide-reaching and that's probably beyond the scope of this PR.

src/templates/post.jsx
```diff
@@ -32,18 +33,87 @@ import {
 } from '../components';
 import { toNodes } from '../util/graphql';
 
+const PostByline = styled.span`
+  font-size: 1.125em;
```
kalongthewires: Should the `em()` function be used in this file for all these `em` values?

src/templates/post.jsx
```diff
@@ -32,18 +33,87 @@ import {
...
+
+  ${mediaQuery.smedium`
+    font-size: 1.5em;
+  `};
+`;
+
 const PostHeader = styled(Header)`
-  margin-top: 4em;
+  align-items: center;
+  background-color: #000;
```
kalongthewires: Should use the black color variable from the theme?

src/templates/post.jsx
```diff
@@ -32,18 +33,87 @@ import {
...
+  background-image: url('/img/backgrounds/space-fog-purple.png'),
+    url('/img/backgrounds/star-field.png');
+  display: flex;
+  font-size: 16px;
```
kalongthewires: The font-size should be in ems or rems to be consistent with the rest of the code base.

src/templates/post.jsx
```diff
@@ -32,18 +33,87 @@ import {
...
+  height: auto;
+  margin-top: 3em;
+  padding-bottom: 1.75em;
+  padding-top: 1.75em;
+
+  ${mediaQuery.medium`
+    min-height: 530px;
```
kalongthewires: Is a min-height needed?

src/templates/post.jsx
```diff
@@ -32,18 +33,87 @@ import {
...
+  `};
+`;
+
+const PostHeaderInfo = styled.div`
+  flex: 1;
+
+  ${mediaQuery.large`
+  flex: 3;
+`};
```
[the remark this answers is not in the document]
kalongthewires: Also not sure that the flex: 1 and flex: 3 are needed?

src/components/molecules/PostExcerpt.jsx
```diff
@@ -22,32 +22,29 @@ const Excerpt = styled(P)`
...
-    );
-  }
+function PostExcerpt({ author, date, excerpt, path, title }) {
+  return (
+    <PostListItem>
+      <PostTitle>
+        <Link to={path}>{title}</Link>
+      </PostTitle>
+      <PostMeta>
+        <Time iso={date} />— {author}
```
kalongthewires: Looks like a space is needed before the dash.

kalongthewires: Author is not required but there's no conditional logic here so the dash will always show up.

src/components/molecules/GlobalNavigation.jsx
```diff
@@ -246,7 +246,7 @@ class GlobalNavigation extends React.Component {
     return (
       <NavBar
         className={
-          this.props.windowWidth <= 480 && this.state.isOpen
+          this.props.windowWidth <= 764 && this.state.isOpen
```
kalongthewires: Should probably put 764 into a named constant so it's not just a magic number.

src/templates/post.jsx
```diff
@@ -88,34 +158,39 @@ const markdown = raw => (
...
+function PostTemplate({ author, children, date, title }) {
+  return (
+    <Main>
+      <GlobalNavigation />
+      <PostHeader>
+        <PostHeaderInfo>
           <PostTitle>{title}</PostTitle>
           <PostMeta>
-            <Time iso={date} /> — {author}
+            {author && <PostByline>Posted by {author}</PostByline>}
```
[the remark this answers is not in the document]
MLKendall: I can go either way, but I did mean this to be `Byline`.
https://www.merriam-webster.com/dictionary/byline

src/templates/post.jsx
```diff
@@ -88,34 +158,39 @@ const markdown = raw => (
...
+            </PostTime>
           </PostMeta>
-        </PostHeader>
-        <PostContent>{children}</PostContent>
-      </Main>
-    );
-  }
+        </PostHeaderInfo>
+        <PostHeaderImage>
+          <img src="/img/monitor-constellation.png" alt="" />
```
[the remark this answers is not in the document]
MLKendall: At some point I think we want to use the different images to mean different things, but until we implement categories, I think they are all going to have the same image for all blog posts.  So, this is just decorative, then, right?  Do we need an alt for that?

src/util/style.js
MLKendall: I will make a separate issue for this.

src/templates/post.jsx
MLKendall: Yes?  I wanted the headers to look consistent regardless of the length of the title on larger screens, but allow it to grow if we had an extraordinarily long title.

src/components/molecules/GlobalNavigation.jsx
```diff
@@ -19,6 +19,8 @@ const mobileBottomMenu = {
   value: `${em('75px')}`,
 };
 
+const MOBILE_NAVIGATION_BREAKPOINT = `767`;
```
khamper-nj: Switch to normal quotes: "767"?

src/templates/post.jsx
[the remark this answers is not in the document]
khamper-nj: Ah I gotcha. My B.

src/templates/post.jsx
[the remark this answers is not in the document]
khamper-nj: That's fair. I try to put decorative images in the CSS as backgrounds instead. Not necessarily required though. I'm cool with leaving it as-is.

src/templates/post.jsx
khamper-nj: Fair enough. I tend to use padding + content to define height and keep things centered without flex, etc. but I see what you mean and this works just as well I think!

PR #369, opened 2019-02-05
feat(jobs): Create Jobs Description Page

### LINKED ISSUE #337: Create Jobs Description Body (Organism)
Create the Jobs Description Page Body (Organism)

Similar to the Blogs, needs to dynamically load the data for the given job.

jobs_description_body_organism

* https://www.figma.com/file/KFeaA4xNivQ7t9xvrNrjkQ/Jobs-For-Backlog?node-id=0%3A1

### LINKED ISSUE #338: Create Jobs Description (Page)
Assemble the Jobs Description Page.

Prereqs:
- [ ] #326 Header
- [ ] #337 Body

Eventually the content can be dynamically included, but that is not necessary to complete this Issue.

jobs_description

* https://www.figma.com/file/KFeaA4xNivQ7t9xvrNrjkQ/Jobs-For-Backlog?node-id=0%3A1

sludemann:
Closes #338
Closes #337


_content/jobs.json
```diff
@@ -1,14 +1,53 @@
...
-    "url": "#"
+    "position": "Open Call",
+    "description": "We don't currently have any specific needs but if you're interested in us then we're interested in hearing from you.",
+    "sections": [
+      {
+        "id": 1,
+        "header": "Hallmarks of a Base Two Employee",
+        "bullets": [
+          "We value self-improvement and keeping up on trends and best-practices.",
+          "We strive to communicate clearly and respectfully with clients and teammates in both in-person and remote settings.",
```
reeddunkle: Maybe:
> We strive to communicate clearly and respectfully with everyone, in-person and remote.

reeddunkle: I wonder if we should try using the "hard" verbs for these, something like:

> * We communicate clearly and respectfully with everyone, in-person and remote.
> * We collaborate, working together to build the best solutions for our clients, for other developers, and for the end-users.
> * We're confident in our abilities, curious about what we don't know, and we share our knowledge.

_content/jobs.json
```diff
@@ -1,14 +1,53 @@
...
+          "We appreciate a teamwork mentality, where all are willing to work together to come to the best solution for any given situation.",
+          "We strive to be confident in our abilities, curious about what we don't know, and willing to share our knowledge with others.",
+          "We are driven to create the best solutions, not just for our clients, but also for other developers and end users."
```
reeddunkle: This one is kinda similar to the third bullet ("create best solutions"), and can possibly be merged with that one.

sludemann:
For other's context the content is based on what we have in the [Basic Job Description](https://docs.google.com/document/d/1WGb7i-tdxT99nBjS7vG76SNd3VhT6nrrO6JRgRgzi30/edit#) that's been shared a few times.  The only thing I've left out so far is the Interview Process because I think that's a littler shakier.  Might be worth keeping the content between the two in sync as we give feedback here and make edits?

_content/jobs.json
sludemann: Maybe, "We strive to be confident in our abilities, curious about what we don't know, and willing to share our knowledge to create the best solutions as a team." ?

src/components/organisms/jobs/JobDescription.jsx
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  &:last-of-type {
+    margin-bottom: ${rem('128px')};
+  }
+`;
+
+const SubHeading = styled(H3)`
+  color: ${themed('color.grey')};
+  font-size: 1.75em;
```
kalongthewires: Use `rem()` so it's clear what the intended font-size was at a glance.

src/templates/job.jsx
```diff
@@ -0,0 +1,58 @@
...
+} from '../components';
+import { darkTheme } from '../theme';
+
+function JobTemplate({ job }) {
+  return (
+    <ThemeProvider theme={darkTheme}>
+      <Main>
+        <GlobalNavigation />
+        <JobsHeader small />
+        <JobDescription job={job} />
```
kalongthewires: Spread the `job` object onto JobDescription instead so you don't have to reference props inside JobDescription with `job.`. That way the JobDescription takes whatever props it specifically wants, instead of caring about how they're packaged (in an object) outside the component. Tends to make for less refactoring later as well.

src/components/organisms/jobs/JobDescription.jsx
```diff
@@ -0,0 +1,77 @@
...
+const ApplyButton = styled(Link)`
+  ${themed('button.default')};
+  display: block;
+  margin-top: 1em;
+  text-align: center;
+  width: ${em('288px', '24px')};
+`;
+
+function JobDescription({ job }) {
+  const { sections } = job;
```
kalongthewires: See comment below about spreading props into JobDescription rather than having it receive `job`.

src/components/organisms/jobs/JobDescription.jsx
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    <React.Fragment>
+      <JobsSection>
+        <H2>{job.position}</H2>
+        <P>{job.description}</P>
+      </JobsSection>
+      {!isEmpty(sections) &&
```
kalongthewires: lodash `map` already guards for this case so you should be able to omit the `isEmpty` check

src/components/organisms/jobs/JobsContent.jsx
```diff
@@ -110,7 +110,11 @@ function JobsContent({ jobs }) {
               positions:
             </P>
             {map(jobs, job => (
-              <JobExcerpt heading={job.position} key={job.id} url={job.url}>
+              <JobExcerpt
+                heading={job.position}
+                key={job.id}
+                url={`/jobs/description/${job.id}`}
+              >
                 {job.description}
```
kalongthewires: Looks like `url` is still included in the propTypes for JobsContent which is throwing a console error.

_content/jobs.json
```diff
@@ -1,14 +1,67 @@
...
-    "id": "2",
-    "position": "Developer",
-    "description": "Elegant user interface design is an integral part of our work. Using a variety of methods including (but not limited to) user interviews, wireframes, and prototypes, we ensure that the end user’s experience is efficient and enjoyable.",
-    "url": "#"
+    "position": "Open Call",
+    "description": "We don't currently have any specific needs but if you're interested in us then we're interested in hearing from you.",
+    "sections": [
+      {
+        "id": 1,
+        "header": "Hallmarks of a Base Two Employee",
```
kalongthewires: `header` means something different in HTML than `heading` so I'd stick with `heading` here.

_content/jobs.json
```diff
@@ -1,14 +1,67 @@
 [
```
kalongthewires: Did you consider going with yaml files + markdown like in the `posts` folder? Not sure if there are pitfalls to that approach but writing out job postings and displaying them would be simpler.

sludemann: I hadn't, instead just ran with the existing `jobs.json` as a constraint, but didn't see any reason not to switch so I've moved it over to that setup.

src/templates/job.jsx
```diff
@@ -0,0 +1,135 @@
...
+        pre: Pre,
+        strong: Strong,
+        ul: UL,
+      },
+    })
+    .processSync(md).contents;
+
+const markdown = raw => (
+  <HoistChildren>{markdownToElement(grayMatter(raw).content)}</HoistChildren>
+);
```
kalongthewires: How about moving the markdown processing stuffs to a util file to avoid having to maintain this stuff in two spots?

sludemann: Added a `templates.jsx` util file to consolidate the markdown pieces.

PR #392, opened 2019-04-18
chore(site): develop v2 rebase master

chriswhawkins:
Includes all completed/merged work from `develop-legacy` and `develop-merge-in-master`, rebased on the latest changes made to `master`, minus the files related to the in-progress jobs page/work, and a pinch of 🌶, simmered for at least 48 hours over developer tears.

Next steps after this is approved/merged:
-Create a new jobs feature branch and commit the job files that were omitted here
-Rebase the case study feature work against updated develop


.eslintrc
```diff
@@ -8,9 +8,13 @@
   },
   "parser": "babel-eslint",
   "rules": {
-    "jsx-a11y/anchor-is-valid": ["error", {
-      "components": ["Link"],
-      "specialLink": ["to"]
-    }]
+    "react/prefer-stateless-function": "off",
```
MLKendall: I assume this is here to keep it from preventing commits?  There is a card to refactor our components to stateless functional components where needed.

src/components/organisms/index.js
```diff
@@ -1,6 +1,8 @@
+import Clients from './Clients';
```
MLKendall: 🤔 Was using only exports not working?  I think it was originally written that way because it is more concise.  This is a small file, so maybe it doesn't matter much, but just curious why this was changed from  the previous version.

.eslintrc
mewdriller: ☝️ 👍

src/components/organisms/index.js
mewdriller: Agreed. If we're doing `import X from './X';` and `export { X };` then the shorthand `export X from './X'` is preferred.

MLKendall:
The orbits are off-center on small screens (this is not occurring on the live site).  Don't know if we should fix this here or spin it off as another card.
image

src/html.jsx
```diff
@@ -1,109 +1,80 @@
...
+          type="image/png"
+        />
+        <link color="black" href="/img/safari-pinned-tab.svg" rel="mask-icon" />
+        <link
+          href="https://fonts.googleapis.com/css?family=Roboto:100,100i,400,500|Anonymous+Pro"
+          rel="prefetch"
+        />
+        <title>Base Two</title>
+        {headComponents}
+        <script src="https://cdn.polyfill.io/v2/polyfill.min.js" />
```
mewdriller: There's [a polyfill pattern](https://github.com/gatsbyjs/gatsby/issues/10435#issuecomment-446627549) in Gatsby using `gatsby-browser.js`. That method also has the advantage of being able to sniff the client to see if the polyfill is actually needed and then use dynamic `import()` to only pay the cost if we actually need it.

.eslintrc
MLKendall: Actually, after further review, looks like that refactoring was done in the this branch.  So, do we want this to be turned off?

src/html.jsx
```diff
@@ -1,109 +1,80 @@
...
+  htmlAttributes,
+  postBodyComponents,
+  preBodyComponents,
+}) {
+  /* eslint-disable react/no-danger */
+  return (
+    <html lang="en" {...htmlAttributes}>
+      <head>
+        <meta charSet="utf-8" />
+        <meta httpEquiv="x-ua-compatible" content="ie=edge" />
```
mewdriller: 🤔 Looks like we don't have the `react/jsx-sort-props` eslint rule on for this repo. We've been using that on other projects. It might be worth adding as it can help reduce diff-churn in some cases.

src/pages/blog/index.jsx
```diff
@@ -92,6 +104,7 @@ export const pageQuery = graphql`
     }
     posts: allMarkdownRemark(
       sort: { fields: [fileAbsolutePath], order: DESC }
+      filter: { frontmatter: { path: { ne: null } } }
```
mewdriller: 😨 Did we have some invalid posts creeping into the build somehow?

src/pages/index.jsx
```diff
@@ -12,49 +14,37 @@ import {
...
   Technologies,
 } from '../components';
+import theme, { darkTheme } from '../theme';
 import { toNodesWithImage } from '../util/graphql';
+import { mediaQuery } from '../util/style';
 
-class IndexPage extends React.Component {
-  static defaultProps = {};
+// eslint-disable-next-line no-unused-expressions
+injectGlobal`
```
mewdriller: 💭 If we move up to `styled-components@^4.2.0` there is a [new global style mechanism](https://www.styled-components.com/docs/api#createglobalstyle) with some performance/readability advantages.

src/util/templates.jsx
```diff
@@ -0,0 +1,108 @@
...
+  img: PostImg,
+  li: LI,
+  ol: OL,
+  p: PostText,
+  pre: Pre,
+  strong: Strong,
+  ul: UL,
+};
+
+const HoistChildren = props =>
```
mewdriller: ⚠️ This is not needed now that we have `<React.Fragment>` / `<>`.

src/util/templates.jsx
```diff
@@ -0,0 +1,108 @@
...
+  remark()
+    .use(remarkReact, {
+      remarkReactComponents: {
+        ...defaultComponentMap,
+        ...componentMapOverrides,
+      },
+    })
+    .processSync(md).contents;
+
+const markdown = (raw, componentMapOverrides) => (
```
mewdriller: 💭 Now that [MDX](https://www.gatsbyjs.org/docs/mdx/) support is included in Gatsby, we should be able to use the [`components`](https://www.gatsbyjs.org/docs/mdx/customizing-components/) map on the `MDXProvider` to do this in a more conventional way.

src/pages/blog/index.jsx
```diff
@@ -78,6 +82,14 @@ function mapPropsToProps({ data }) {
...
+  posts: PropTypes.arrayOf(
+    PropTypes.shape({
+      id: PropTypes.node.isRequired,
+    }),
+  ).isRequired,
+};
+
 export default mapProps(mapPropsToProps)(BlogIndex);
 
 export const pageQuery = graphql`
```
mewdriller: 🌈 There's a new API for integrating data from GraphQL queries into Gatsby pages/components using [`<StaticQuery />`](https://www.gatsbyjs.org/docs/static-query/). Not something we need to use in this instance, but I think it was worth mentioning.

src/components/atoms/Img.jsx
```diff
@@ -12,73 +12,67 @@ const srcProps = obj =>
     ['src', 'srcSet'],
   );
 
-class Img extends React.Component {
-  static defaultProps = {
-    resolutions: null,
-    sizes: null,
-  };
+function Img({ alt, resolutions, sizes, ...rest }) {
```
mewdriller: 💡 The `gatsby-image` component has improved a lot since we started this project. There's a strong change it covers our use cases more effectively now. It might be worth investigating in another card.

src/components/atoms/index.js
```diff
@@ -1,25 +1,50 @@
+import A from './A';
+import Address from './Address';
+import Blockquote from './Blockquote';
+import Blurb from './Blurb';
+import Button from './Button';
+import CallToAction from './CallToAction';
+import Code from './Code';
+import Del from './Del';
+import EM from './EM';
```
mewdriller: 🤔 Should this be `Em` and `./Em` since it stems from "emphasis"?

src/components/organisms/index.js
tborres: Since the `import X from '.X';` exists many places in the codebase I suggest that we split off a separate tech debt task for this. 

I'll create the card, but I want to confirm the proposed direction. When I quickly make the change to the file to be the following I have tons of red squiggles.

image

Is this the proposed format? (@mewdriller @MLKendall )

tborres:
@MLKendall I was unable to reproduce the orbit issue on small screen (https://github.com/b2io/base2.io/pull/392#issuecomment-484910707) Is this still an issue on this branch, and if so, how can I recreate it? Thanks!

src/html.jsx
tborres: A separate task (#393) has been created for this suggestion so that this PR can move forward.

MLKendall:
Hi Tony:

Yes, still an issue as of this morning.  To reproduce the issue:
- Open in Chrome
- Open Developer Tools
- Preview in iPhone 6/7/8 mode

Let me know if you still can't reproduce it.

*Mandy Kendall*
*Developer*
mkendall@base2.io

*Base Two*
*We turn ideas into beautiful software.*
*base2.io* <http://base2.io>

src/html.jsx
tborres: Task #394 created.

src/pages/index.jsx
tborres: Task #395 created

src/util/templates.jsx
tborres: Task #396 created

src/components/atoms/Img.jsx
tborres: Task #397 created

.eslintrc
tborres: I'll remove the line and give it a whirl.

Issue #397, opened 2019-05-02
Investigate the use of gatsby-image


tborres:
💡 The `gatsby-image` component has improved a lot since we started this project. There's a strong change it covers our use cases more effectively now. Investigate its potential usage and create chore cards as needed.

_Originally posted by @mewdriller in https://github.com/b2io/base2.io/pull/392_

PR #407, opened 2019-06-04
fix: image load jumping

### LINKED ISSUE #404: Image load jumping
The current implementation of `Img.jsx` causes images to be replaced by a `div` anytime they are not in view (`if (!inView) return <div {...rest} ref={ref} />;`) . This may make sense for initial load, but I question whether it makes sense to remove an already loaded image. Furthermore, because the usages of `Img` do not have set height/widths the UI jumps around when the image is added/removed.

Proposed changes
* [ ] lazy load the images in a way that doesn't remove them after they've been loaded
* [ ] maintain the site flow when images are lazy loaded

**Dev Notes**
Consider #397

tborres:
The website appears to jump as the user scrolls down and up because the
image pop in and out as they leave the view (and the replacement doesn't
retain the size). This PR removes the inview logic as a temporary
solution to improving the usability. We'll want to consider rewriting
the logic (per #397) to improve the load times.

Closes #404


src/util/helpers.js
```diff
@@ -10,7 +10,7 @@ const renameKeys = (obj, pairs = []) =>
   }, obj);
 
 const supportsWebp = memoize(() => {
-  if (isUndefined(window)) return false;
+  if (typeof window === 'undefined') return false;
```
mewdriller: 🤔 Is this to reduce bundle size?

src/components/atoms/Img.jsx
```diff
@@ -13,40 +12,23 @@ const srcProps = obj =>
   );
 
 function Img({ alt, resolutions, sizes, ...rest }) {
-  return (
-    <Observer>
```
mewdriller: ⚠️ Removing the `<Observer>` code here will have perf implications as all of the images will load immediately instead of waiting until they're scrolled into view. 

We're not currently leveraging `gatsby-image` at all in this component. Previously that had been done because it was not quite baked and ready, but that's no longer the case.

There are a lot of potential performance improvements we'll get "for free" by using that component effectively.

src/util/helpers.js
tborres: `isUndefined` doesn't work when doing `npm run build` (after removing the `Img` `InView` logic. The `typeof` approach is in line with Gatsby's [documentation](https://www.gatsbyjs.org/docs/debugging-html-builds/).

src/components/atoms/Img.jsx
tborres: I assumed that there would be performance implications, but honestly I think that the usability jumpiness outweighs the performance implication. I had run light house and didn't notice significant differences, but honestly I'm not that familiar with the reports. I had looked into `gatsby-image`, but I'm too unfamiliar with Gatsby, images, and our current image approach to put it in place.

tborres:
@mewdriller are these polyfill changes kosher? If so then I think that this is ready to merge.

mewdriller:
I would back out the await/import change for now unless you tested. I think I might have a better long-term option with a plugin. But removing the IntersectionObserver polyfill is perfect.

src/util/helpers.js
mewdriller: 🤔 That's _super_ weird as [`_.isUndefined()`](https://github.com/b2io/base2.io/pull/407#issuecomment-499221561) is essentially exactly those lines of code. Are any of the other imports from `lodash` breaking? That seems odd that only this one would.

tborres: I don't understand it either. We had run into this issue previously in the `GlobalNavigation.jsx` and switching to the verbatim approach worked there too.

PR #412, opened 2019-06-05
feat(blog): add feature images to blog articles

tborres:


tborres:
While doing the case studies I realized that we could do feature images for blog articles; however, I wasn't sure if we wanted to go in that direction so I split it off into a separate PR to get some feedback.

tborres:
I'm going to merge this in after getting approval verbally from Caitlin on the concept.

PR #417, opened 2019-08-02
fix(jobs): limit resume file size

### LINKED ISSUE #415: Open positions max width
The Open Positions area of the jobs page should have the same max-width as the rest of the page.

image

### LINKED ISSUE #416: Resume submission limit
We use https://getsimpleform.com/ for our form submissions. There appears to be a file size limit that results in the user being redirected to a 413 error message. After a bit of testing it looks like the limit is somewhere at or below 1.5mb, but I'm not sure of what the exact limit is.

Recommendation:
- [x] Try adding `enctype="multipart/form-data"` per http://docs.liveformhq.com/#file-uploads and testing again
- [x] Update the form to specify a 1mb file limit
- [x] Try to prevent the user from uploading a file larger than 1mb

tborres:
- update label to specify limit
- add validation messaging
- reduce width of open positions

Closes #416
Closes #415


src/components/organisms/jobs/JobForm.jsx
```diff
@@ -126,10 +128,21 @@ function JobForm({ isApprenticeship, position }) {
...
+              label="Please upload your resume (<1MB)"
               type="file"
               name="resume"
-              multiple
+              onChange={e => {
+                if (
+                  e.target.files &&
+                  e.target.files[0].size > FILE_SIZE_LIMIT
+                ) {
+                  e.target.setCustomValidity(
```
mewdriller: ⚠️ We should generally avoid using imperative DOM methods like this. It's possible that React will recycle/replace the underlying node out from under us between renders.

This could be done using component state and a ref in a way that should make it safe regardless of the decisions the React reconciler makes about keeping/replacing nodes. It would also probably make sense to encapsulate that into a `FileInputField` component with a `maxSize` prop to make this form component easier to scan.

Alternatively, you could use a styled DOM node to render the error text ourselves and avoid needing to manage the ref and keep the imperative API in sync.

src/components/organisms/jobs/JobsContent.jsx
```diff
@@ -71,7 +71,7 @@ function JobsContent({ jobs }) {
           <LI>Fancy office coffee</LI>
         </UL>
       </Section>
-      <Section id="positions">
+      <Section id="positions" narrow>
```
mewdriller: 🌟

src/components/atoms/form/TextField.jsx
```diff
@@ -46,11 +46,14 @@ class TextField extends React.Component {
...
-    this.setState(stateUpdateFromValue(event.target), this.notifyAfterChange);
+    this.setState(
+      stateUpdateFromValue(event.target),
+      this.notifyAfterChange(event),
+    );
   };
 
-  notifyAfterChange = () => {
-    this.props.onChange(this.state.value);
+  notifyAfterChange = event => {
```
mewdriller: 🤔 It looks like `notifyAfterChange` is only called from within `handleChange`. It'll probably be clearer if this behavior is just inlined into that function body.

src/components/organisms/jobs/JobForm.jsx
tborres: This makes sense to me in theory, but I'm not sure of how to properly encapsulate the behavior into a new `FileInputField` component.

PR #421, opened 2019-08-02
fix: ie11 image not appearing

### LINKED ISSUE #418: IE11 image not appearing
The blue astronaut on the Jobs page header is not displaying in IE11. It appears that IE11 is showing as accepting webp files, but it isn't displaying the image.

image

tborres:
Went with the straightforward path of just using the png image rather
than trying the webp.

Closes #418

### DIFF
--- a/src/components/organisms/jobs/JobsHeader.jsx
+++ b/src/components/organisms/jobs/JobsHeader.jsx
@@ -4,7 +4,6 @@ import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import { mediaQuery, themed } from '../../../util/style';
 import { A, BaseTwoJobsLogo, Header, P } from '../../atoms';
-import { supportsWebp } from '../../../util/helpers';
 
 const imageSize = {
   default: '260px',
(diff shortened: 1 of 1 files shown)


src/components/organisms/jobs/JobsHeader.jsx
```diff
@@ -91,15 +90,11 @@ const JobsButton = styled(A)`
...
-const astronautSource = supportsWebp
-  ? `/img/jobs/BlueAstronaut.webp`
-  : '/img/jobs/BlueAstronaut.png';
-
 function JobsHeader({ small }) {
   return (
     <StyledHeader>
       <AstronautImg
-        src={astronautSource}
+        src="/img/jobs/BlueAstronaut.png"
```
mewdriller: 🤔 Ideally we'd be leveraging something like `gatsby-image` to do the heavy lifting on the image optimization for us. This will definitely work until we get that wider image support landed.

tborres: For reference: #397

Issue #386, opened 2019-04-15
Create Footer


MLKendall:
Create a new footer for use on the Case Studies List and Individual Case Studies pages, based on the [Figma Design](https://www.figma.com/file/bJsG3hEp3zsDdKgZJARPWxL7/Case-Studies?node-id=1%3A2)

![image](https://user-images.githubusercontent.com/8663544/56159214-1b561100-5f92-11e9-91ec-6ddc63324f4f.png)

Acceptance Criteria
- [x] Footer appears on Case Studies List and Individual Case Studies pages
- [x] `Get Started` button links to the Contact section of the home page.
- [ ] Works and renders properly on Desktop, Tablet and Mobile
- [ ] Test on all appropriate browsers (Chrome, Firefix, IE11, Edge, Safari)

Story: #382

Issue #382, opened 2019-04-15
Case Study MVP Implementation


MLKendall:
Implement pages related to Case Studies MVP

Implementation should be based on the [Figma Design](https://www.figma.com/file/bJsG3hEp3zsDdKgZJARPWxL7/Portfolio-Items?node-id=1%3A2)

Acceptance Criteria:

- [x] #383  
- [x] #384 
- [ ] #385
- [ ] #386 
- [ ] #387 
- [x] #389 
- [x] #390

PR #429, opened 2019-10-30
feat: Case Studies Footer

k8hansen:
Closes: #386
Story: #382 

This differs from the comp on [Figma](https://www.figma.com/file/bJsG3hEp3zsDdKgZJARPWxL7/Case-Studies?node-id=1%3A2) as a collective decision was made to drop the left-justified text and instead center a button linking to contact info.
 
footer

TODO
- [ ] Testing in I11 and Edge.


src/components/atoms/CallToActionLink.jsx
```diff
@@ -0,0 +1,22 @@
...
+const CallToActionLink = styled(GatsbyLink)`
+  ${themed('button.default')};
+  display: inline-block;
+  padding: ${em('10px')} ${em('13px')};
+  text-decoration: none;
+  text-transform: capitalize;
+
+  ${mediaQuery.small`
+  font-size: 1.875em;
+`};
```
reeddunkle: 🎨 Should indent these two lines to create a visual block. Ideally our VS Code formatter would do this for us, but it gets caught up on styled-components syntax sometimes.

src/components/atoms/CallToActionLink.jsx
```diff
@@ -0,0 +1,22 @@
...
+
+  ${mediaQuery.large`
+  font-size: 2.625em;
+`};
```
reeddunkle: 🎨 Indent

src/components/organisms/caseStudies/CaseStudyFooter.jsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import styled from 'styled-components';
+import { Button } from '../../atoms';
+import { mediaQuery, webpBackground } from '../../../util/style';
+import { em } from 'polished';
+import { CallToActionLink } from '../../atoms';
+
+const Wrapper = styled.div`
+  align-items: flex-start;
+  padding-top: 4vh;
```
reeddunkle: 🎨 Let's order these alphabetically

src/templates/caseStudy.jsx
```diff
@@ -13,6 +13,7 @@ import {
   GlobalNavigation,
   P,
   Section,
+  CaseStudyFooter,
```
reeddunkle: 🎨 Let's order these alphabetically

src/components/atoms/CallToActionLink.jsx
```diff
@@ -0,0 +1,22 @@
+import GatsbyLink from 'gatsby-link';
+import styled from 'styled-components';
+import { mediaQuery, themed } from '../../util/style';
+import { em } from 'polished';
+
+const CallToActionLink = styled(GatsbyLink)`
```
reeddunkle: 🤔 Another way to tackle this problem would be to create this `CallToActionLink` component inside the already existing `CallToAction.jsx` file. That file could export both the `CallToAction` (anchor) component and your new `CallToActionLink` component. If they're in the same file, then their shared styles could be lifted into a common variable. That way if we want to change the way a "CallToAction" looks on the site, we only have to change it in one place and it will apply to both components.

That said, the way it exists here is fine. It's totally reasonable to decide to keep them separate, and we can revisit the idea of collocating them in the future.

k8hansen: Understood, I think this is what I was conceptually trying to do, although I was trying to do it in the footer file. 

However, without knowing too much about the site, this is my line of thinking - I'm not sure that there are only two buttons.  If there were multiple buttons I would assume that they would be put in one place.  Being that there is only one component in that call to action file, I would then assume that all other buttons are separated.  

Does that make sense? It is also possible that I'm not understanding react components fully.  All that aside, I think there was a sense of urgency to get this out, so if this is an acceptable solution I support leaving it as-is.

src/components/atoms/CallToActionLink.jsx
k8hansen: Done. Can you explain why you changed the title of the PR?  What is a CTA?

src/components/atoms/CallToActionLink.jsx
chriswhawkins: If I'm understanding correctly, it looks like we should have been using the `gatsby-link` component already in some of the existing use cases, so perhaps this is a good opportunity to go ahead and update the existing `CallToAction` component to automatically return the correct `Link` or `a` tag. See: https://www.gatsbyjs.org/docs/gatsby-link/#reminder-use-link-only-for-internal-links

k8hansen: @chriswhawkins would this be an example of a change I should make on this PR, or open a new issue for? Not sure if I'm understanding correctly and we should go back and change to the link component and that needs to be a separate issue.

src/components/organisms/caseStudies/CaseStudyFooter.jsx
```diff
@@ -0,0 +1,35 @@
...
+  return (
+    <Wrapper>
+      <CallToActionLink to="/#contact-us">Get Started</CallToActionLink>
+    </Wrapper>
+  );
+}
+
+CaseStudyFooter.defaultProps = {};
+
+CaseStudyFooter.propTypes = {};
```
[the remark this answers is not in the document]
mewdriller: ☝️ We could consider just stripping the `propTypes` throughout the project in a tech-debt card as they're not providing value at this point.

src/components/organisms/caseStudies/CaseStudyFooter.jsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import styled from 'styled-components';
+import { Button } from '../../atoms';
+import { mediaQuery, webpBackground } from '../../../util/style';
+import { em } from 'polished';
```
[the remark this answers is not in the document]
mewdriller: 🤔 Maybe we should add a card to setup `husky` and `lint-staged` and/or CI on the repository?

cantyjeffrey:
Hey @k8hansen is this PR still a thing?

k8hansen:
@cantyjeffrey YES!  If you would like to take it from here please be my guest!!

zachcd:
This renders in IE11 and Edge for me.  Gatsby develop had issues but build and serve worked for me.

src/components/organisms/caseStudies/CaseStudyFooter.jsx
```diff
@@ -0,0 +1,35 @@
...
+import { CallToActionLink } from '../../atoms';
+
+const Wrapper = styled.div`
+  align-items: flex-start;
+  background: no-repeat bottom;
+  background-size: cover;
+  display: flex;
+  justify-content: center;
+  margin-top: 6vw;
+  min-height: 45vw;
```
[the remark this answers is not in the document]
zachcd: Good for maintaining aspect ratio.

Issue #435, opened 2020-01-15
Email validation


tborres:
Our email validation logic (on contact us and job postings) doesn't allow the first character of the email address to be capitalized even then capital letter should be accepted.

zachcd:
#436 closed

PR #436, opened 2020-01-21
Validate capital letters in email form entry

zachcd:
Add A-Z into the capture groups of the email regex.


tborres:
Closes #435

reeddunkle:
I tested too. Thanks for this 💃

Issue #282, opened 2018-03-05
Site Does Not Scroll Once Navigation Item is Picked


MLKendall:
When a navigation item is chosen on mobile, you can no longer scroll the site.  The link also only goes to the designated section on second click of the nav button.

Device / Broswer Tested: iPhoneSE / Chrome.

csteinert:
![282-scroll-issue](https://user-images.githubusercontent.com/1918353/37048175-14e4d8c4-213b-11e8-898b-e720a98e0aa4.gif)

Issue #251, opened 2018-02-28
Mobile Navigation Does Not Close After Link is Clicked


MLKendall:
The mobile navigation menu currently does not close after a link is clicked.  Scrolling to the top of the page reveals the still-open menu.  

Desired Behavior:  Navigation should close when a link is clicked (except for the phone and email buttons)

Devices Tested: iPhone SE, Google Pixel (original)

csteinert:
Got some video of this bug:

![251-mobile-nav-issue](https://user-images.githubusercontent.com/1918353/36801006-e5760178-1c7e-11e8-8202-0ac0072c526e.gif)

PR #448, opened 2020-02-19
feat(site): Smooth scrolling on nav link

### LINKED ISSUE #237: Smooth Scroll for Navigation
Currently, the navigation jumps abruptly to a new section.  To create a better user experience, implement a smoother scroll to each of the individual sections.

- [x] Research the best method to implement a smooth scroll

- [ ] Implement the smooth scroll

- [ ] Test (include https://gist.github.com/csteinert/6c02cc6f17329c754e15fc4e2e762dc6 in PR)

https://css-tricks.com/smooth-scrolling-accessibility/

**Notes**
Current issues with the implementation (https://github.com/fisshy/react-scroll) (branch: 237-smooth-scroll):
- Scrolling links do not have proper hover cursor styles
- Global Nav uses the same markup on the Blog pages, but the scroll links only work if they're on the same page as the linked content (so you can't navigate back)
- The home page jumps back to the top any time the mobile nav panel is opened on small screens.

zachcd:
### feat(site): Smooth scrolling on nav link use
* Add smooth scrolling package that applies to all anchor links
* Change top link to a #top anchor tag

note: the polyfill for this has a conflict with Gatsby SSR, so I'm importing the non-polyfilled script directly.  Browsers without support will stick to jump navigation instead of scroll.

smoothscrollb2io

closes #237

### DIFF
--- a/src/components/molecules/GlobalNavigation.jsx
+++ b/src/components/molecules/GlobalNavigation.jsx
@@ -20,7 +20,9 @@ const mobileBottomMenu = {
   value: `${em('75px')}`,
 };
 
-
+if (typeof window !== `undefined`) {
+  SmoothScroll('a[href*="#"]', {speed: 700,speedAsDuration: true, easing: 'easeOutQuad'});
+};
 
 const MOBILE_NAVIGATION_BREAKPOINT = '767';
 
@@ -247,10 +249,6 @@ class GlobalNavigation extends React.Component {
     this.setState({ isOpen: false });
   };
 
-  componentDidMount() {
-    SmoothScroll('a[href*="#"]', {speed: 700,speedAsDuration: true, easing: 'easeOutQuad'});
-  }
-  
   render() {
     return (
       <NavBar
--- a/src/components/molecules/GlobalNavigation.jsx
+++ b/src/components/molecules/GlobalNavigation.jsx
@@ -4,6 +4,7 @@ import styled from 'styled-components';
 import PropTypes from 'prop-types';
 import { em } from 'polished';
 import windowSize from 'react-window-size';
+import SmoothScroll from 'smooth-scroll';
 import {
   A,
   BaseTwoLogo,
@@ -19,6 +20,8 @@ import { mediaQuery, themed } from '../../util/style';
 const mobileBottomMenu = {
   value: `${em('75px')}`,
 };
+
+SmoothScroll('a[href*="#"]', {speed: 700,speedAsDuration: true, easing: 'easeOutQuad'});
 
 const MOBILE_NAVIGATION_BREAKPOINT = '767';
 
@@ -257,7 +260,7 @@ class GlobalNavigation extends React.Component {
       >
         <MainNavigationContainer>
           <LogoToggleContainer>
-            <HomeLink href="/" title="Base Two - Home">
+            <HomeLink href="/#top" title="Base Two - Home">
               <BaseTwoLogo />
             </HomeLink>
             <MenuToggle aria-label="Toggle Menu" onClick={this.handleClick}>
--- a/src/components/molecules/GlobalNavigation.jsx
+++ b/src/components/molecules/GlobalNavigation.jsx
@@ -3,7 +3,7 @@ import styled from 'styled-components';
 import PropTypes from 'prop-types';
 import { em } from 'polished';
 import windowSize from 'react-window-size';
-import SmoothScroll from 'smooth-scroll/dist/smooth-scroll'
+import SmoothScroll from 'smooth-scroll/dist/smooth-scroll';
 import {
   A,
   BaseTwoLogo,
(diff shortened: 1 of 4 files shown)


zachcd:
looks like there's a deploy issue with the package I used.  Something to work on.

FrankyMartz:
What about the native `<element>.scrollIntoView()`? It looks like it has decent support and a Polyfill.

https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

**Polyfill:**
http://iamdustan.com/smoothscroll/

zachcd:
@FrankyMartz I'm using smooth-scroll for easing, speed, and now offset purposes.  Just a little more control over the animation.

@tborres looks like the nav menu on mobile locks scrolling while its open.  Shouldn't be too hard to fix.

zachcd:
Bringing in Issues and PRs: #251 #261 #282 #285 

The Mobile Nav was resetting scroll position on open and preventing scrolling while open so as to make sure that touch scrolling wouldn't accidentally happen on the background viewport while in the mobile nav.  This disabling of scrolling prevented the smooth-scroll from happening. 

In order to keep the same functionality but allow programmatic scrolling(and solve the resetting of scroll position): I set touch-action: none, see: https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action

src/components/molecules/GlobalNavigation.jsx
```diff
@@ -3,6 +3,7 @@ import styled from 'styled-components';
 import PropTypes from 'prop-types';
 import { em } from 'polished';
 import windowSize from 'react-window-size';
+import SmoothScroll from 'smooth-scroll/dist/smooth-scroll'
```
tborres: ❓ Couldn't you import directly from the npm package name?

```suggestion
import SmoothScroll from 'smooth-scroll';
```

Also, missing a semi-colon. With the right plugins and settings you can setup auto-fix on save so that most of these common misses are auto-fixed.  

**VS Code Extensions**
- ESLint (2.1.1)
- Prettier (3.20.0)

**Settings:**
```
"editor.formatOnSave": true,
"editor.codeActionsOnSave": {
    "source.fixAll.eslint": true
  },
```

src/components/molecules/GlobalNavigation.jsx
```diff
@@ -19,6 +20,16 @@ const mobileBottomMenu = {
   value: `${em('75px')}`,
 };
 
+if (typeof window !== `undefined`) {
+  SmoothScroll('a[href*="#"]',
```
tborres: 🎨 Our eslint/prettier rules would typically put the opening `{` on this line.

src/components/molecules/GlobalNavigation.jsx
```diff
@@ -19,6 +20,16 @@ const mobileBottomMenu = {
   value: `${em('75px')}`,
 };
 
+if (typeof window !== `undefined`) {
+  SmoothScroll('a[href*="#"]', 
+  {
+    easing: 'easeOutQuad', 
+    speed: 700,
+    speedAsDuration: true,
+    offset: '50'
```
tborres: Prefer alphabetical order of properties to reduce the chance of duplicates and ease future merges (though if you look around you can see that this rule isn't always followed).

src/components/molecules/GlobalNavigation.jsx
```diff
@@ -34,6 +45,7 @@ const NavBar = styled.nav`
   &.menu-open {
     height: 100vh;
     transition: height 300ms ease-in;
+    touch-action: none;
```
tborres: Alphabetical order

src/pages/index.jsx
```diff
@@ -32,7 +32,7 @@ injectGlobal`
   }
 
   .noScroll > div {
-    height: 100vh;
+    height: 100%;
```
tborres: ❓ Why is this change needed?

src/components/molecules/GlobalNavigation.jsx
zachcd: The import from the dist is because the package references window, so when Gatsby imports it to complete a build, it crashes.  This skips the package's wrapper that references window.  If there's a more "gatsby"-like solution, I'd be very happy to hear about it.

tborres: @mewdriller do you know of a more gatsby-like approach to this window issue that Zach ran into?

src/pages/index.jsx
zachcd: 100vh was shortening the height of the div when the menu opened, resetting scrolling position.  This is a bugfix that I'm adding in that fixes the jump to the top on hamburger menu open bug.

Setting the height is still important because of the overflow setting. 
```In order for overflow to have an effect, the block-level container must have either a set height (height or max-height) or white-space set to nowrap.```

zachcd: It was needed before, because the 100vh was resetting the scroll position and making it jump to the top, 100% was fixing that issue.  I can just delete this whole style now as we don't need it with the use of touch-action.

src/components/molecules/GlobalNavigation.jsx
zachcd: https://www.gatsbyjs.org/docs/using-client-side-only-packages/

They don't have anything specific, and their first recommendation is to "work around the problem"

Issue #385, opened 2019-04-15
Create Header for Individual Case Studies Pages


MLKendall:
Create a header for the Individual Case Study pages, bases on the [Figma Design](https://www.figma.com/file/bJsG3hEp3zsDdKgZJARPWxL7/Case-Studies?node-id=1%3A2)

![image](https://user-images.githubusercontent.com/8663544/56158225-80f4ce00-5f8f-11e9-85ec-009216733930.png)

Acceptance Criteria:
- [ ] Includes the Base Two logo in all instances, as shown
- [ ] Includes a place for the logo for a particular client (where Sherwin-Williams logo is shown in mock).
- [ ] Works on Desktop, Tablet and Mobile
- [ ] Tested on all appropriate browsers (Chrome, Firefox, IE11, Edge, Safari)

Story: #382

PR #458, opened 2020-02-27
feature(case-studies)Case Study Header (List View, Case Study Pages)

FrankyMartz:
## Overview

Add Header to Individual Case Studies. Increase client logo image resolutions for case study pages

## Description

* Create CaseStudy Header Component
* Replace CaseStudy Client images with higher resolution alternatives
* Adjust CaseStudy List Page client images to handle higher resolution images
* Modified `b2-logo-with-text.svg` to remove excess whitespace (IE11)

## Test Plan

Smoke Test conducted on local machine and VMs.

**MacOS**
- [x] Chrome
- [x] Safari
- [x] FireFox

**Microsoft**
- [x] Edge
- [x] IE11

Screen Shot 2020-02-26 at 11 03 48 PM
Screen Shot 2020-02-26 at 11 04 13 PM
Screen Shot 2020-02-26 at 11 04 52 PM
Screen Shot 2020-02-26 at 11 03 30 PM

## Closing Issues
closes. #385


src/components/molecules/ClientHeader.jsx
```diff
@@ -0,0 +1,132 @@
+/* ========================================================================== *
+ * ClientHeader
+ * ========================================================================== */
```
chriswhawkins: I'm not necessarily opposed, but we don't have these pretty comment titles elsewhere 🤷‍♂

src/components/molecules/ClientHeader.jsx
```diff
@@ -0,0 +1,132 @@
...
+import { em } from 'polished';
+
+import { mediaQuery, themed } from '../../util/style';
+import { Header, Img, H1 } from '../../components';
+
+const BGAspectRatio = '2.65';
+
+const ClientHeaderContainer = styled(Header)`
+  display: flex;
+  height: auto;
```
chriswhawkins: We try to alpha-order css/style properties where possible. Linting might not be set up on the project, but maybe we can get a task for adding it.

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@b2io/base2.io",
-  "version": "2.0.0",
+  "version": "2.1.0",
```
tborres: We generally don't include the package-lock file in the PR unless there were changes to the `package.json` file.

src/components/molecules/ClientHeader.jsx
tborres: I'd vote to remove this header comment. We have not done this style of commenting elsewhere in the site or our other projects.

src/templates/caseStudy.jsx
```diff
@@ -104,19 +106,19 @@ const renderMarkdown = raw =>
   });
 
 function CaseStudyTemplate({ caseStudy }) {
-  const { client, highlights, link, project, technologies } = caseStudy;
```
tborres: ❓ Why did you switch away from the full destructing? I'm partially asking because this feels like something that the next developer may switch back if we don't have a sound reason for one way vs another.

src/components/molecules/ClientHeader.jsx
FrankyMartz: Oops, will :fire:

src/templates/caseStudy.jsx
FrankyMartz: At the time, I was exploring and expanded to dot notation because it helped me quickly reference / associate where values were originating while developing – envisioning values as properties on a data blob vs floating variables.

In this case, I personally believe I should revert this to destructured and fully intend to.

> How do you reason object destructuring then?
 
- When properties are referenced more than once (duplicate property lookups are inefficient)
- And when scope is limited to one data structure (enforcing an explicit visual association). Destructuring multiple data structures in a scope means backtracking to data origin for the un-aquianted.

NOTE: This isn't my hard fast rule though. As I believe team consensus wins :100: of the time. Consistency is wayyy more important to code readability in my book.

Btw, thanks for asking! Always happy to expand on the reasoning behind my intentions :)

src/components/molecules/ClientHeader.jsx
FrankyMartz: :fire: removed

src/components/molecules/ClientHeader.jsx
FrankyMartz: ✅ CSS Sorted

package-lock.json
FrankyMartz: ◀️ `package-lock.json` Reverted

src/templates/caseStudy.jsx
```diff
@@ -8,10 +8,12 @@ import { mediaQuery } from '../util/style';
 import markdown from '../util/templates';
 
 import {
+  ClientHeader,
   CaseStudyAside,
   Main,
   GlobalNavigation,
   P,
+  // PageHeader,
```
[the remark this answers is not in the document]
FrankyMartz: 🔥 commented code removed

src/templates/caseStudy.jsx
FrankyMartz: ◀️ Reverted back to destructured

kesslern:
Looks good to me. The only thing I noticed is that some of the image file sizes are a little large. We could improve loading times by converting them to jpegs.

PR #459, opened 2020-02-27
(fix): Team Title collisions

### LINKED ISSUE #451: Team title collisions
With the new team titles we're having collisions at smaller screens (e.g. 500px width)

image

zachcd:
## Remove Team title collisions

Extend Extra-Small screen styles up to Medium screens:

* TeamList
* ListItem
* :After
*  child div

closes #451


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@b2io/base2.io",
-  "version": "2.0.0",
+  "version": "2.1.0",
```
tborres: Generally speaking, if you didn't modify the `package.json` then we don't include the lock file in the PR.

PR #466, opened 2020-03-02
fix(site): Blog header compression on IE11

### LINKED ISSUE #374: Blog - IE11
Title text doesn't adjust as expected in IE11

image

kesslern:
Closes #374

IE11 has some irregular behavior with flexboxes that require explicitly setting the width sometimes.

Before:
Before

After:
After


kesslern:
If anyone is curious, this was tested using Microsoft's virtual machine images they provide for testing with Edge and IE11.

https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/

PR #469, opened 2020-03-05
Case study footer (#386)

kesslern:
This PR adds the footer to the Case Studies page. To avoid duplicating code and to ensure the footer appears the same on the Case Studies page and the main landing page, I pulled the footer into a reusable component. In doing so, I ran across a couple bugs in the Contact Us area that made sense to knock out at the same time.

[Screenshot of the Case Studies footer](https://user-images.githubusercontent.com/7218166/76016428-ca714300-5eea-11ea-92d9-a58651a0bd09.png)

Bug fixes:
- Contact form header `z-index` was incorrectly set and could [overlap the navbar](https://user-images.githubusercontent.com/7218166/76015223-e70c7b80-5ee8-11ea-896f-e2b9f0b5d7d5.png).
- In certain screen sizes, the absolutely-positioned globe would [overlap the footer](https://user-images.githubusercontent.com/7218166/76015415-2e930780-5ee9-11ea-9bf3-9d6526e6f826.png).

Improvements:
- Add another level of z-index to the theme so the outer-space objects like planets and stars can sit between the blackness of space (the black background) and the main content.
- Add ability to style buttons as purple by passing `purple` prop to the Button component ([example](https://user-images.githubusercontent.com/7218166/76016428-ca714300-5eea-11ea-92d9-a58651a0bd09.png))
- Provide optimized `webp` and `jpg` footer images for mobile and desktop
- Better text wrapping in the location section. It's a bit longer since it includes Chicago now. I thought it looked awkward when it was wider than the logo. ([before](https://user-images.githubusercontent.com/7218166/76015803-c4c72d80-5ee9-11ea-9afa-fd4124598e7c.png), [after](https://user-images.githubusercontent.com/7218166/76015879-df99a200-5ee9-11ea-82c1-2022d31f433a.png))
- Increase spacing between the contact form and footer so the submit button doesn't touch the moon's surface, more closely matching the Figma designs. ([before](https://user-images.githubusercontent.com/7218166/76016039-212a4d00-5eea-11ea-81a0-7af6626550ad.png), [after](https://user-images.githubusercontent.com/7218166/76016064-2ab3b500-5eea-11ea-9907-431cae8e3197.png)) (in the before, you can't even see the Google Maps link!)

Props to Franky & Kate for the previously-completed work this was built upon.

Tested on:
* Pixel 2 (Chrome)
* iPad Pro (Safari)
* Safari Mac
* Safari Windows
* Chrome Windows
* Chrome Linux
* Firefox Windows
* IE11


src/components/organisms/caseStudies/CaseStudyFooter.jsx
```diff
@@ -0,0 +1,51 @@
...
+import Footer from '../../molecules/Footer';
+
+const Wrapper = styled.footer`
+  margin-top: 10em;
+`;
+
+const LinkWrapper = styled(GatsbyLink)`
+  width: 328px;
+  margin: 0 auto 10em;
+
```
tborres: Is the spacing between these attributes on purpose? Feels a bit odd to me. We also typically try to have this in alphabetical order (here and elsewhere in this PR).

src/components/molecules/Footer.jsx
```diff
@@ -0,0 +1,47 @@
...
+const footerHeight = '300px'
+
+const Wrapper = styled.footer`
+  position: relative;
+  height: ${footerHeight};
+  width: 100%;
+`;
+
+const Background = styled.div`
+  position: absolute;
```
tborres: We generally shoot for alphabetical order on props

src/theme.js
```diff
@@ -89,22 +90,23 @@ function makeTheme(color) {
 
   const button = {
     default: css`
-      background-color: ${color.black};
+      background-color: ${props => props.purple ? color.purple : color.black};
```
tborres: This approach works, but is a bit limiting if we want other colors in the future. I don't have a better suggestion so keep it as-is unless a quick better architecture springs to mind.

src/components/organisms/caseStudies/CaseStudyFooter.jsx
kesslern: I don't think it was, I'll go through and alphabetize the CSS in the changed sections.

src/theme.js
kesslern: My first approach was a second `styled.button` with some overridden CSS, but the theme's colors aren't exported from the theme so I ended up going with this approach. If we wanted to support more button colors in the future, I think we could change this to a `color` prop that takes a string, similar to Material UI's `variant` prop, and then include that color's specific CSS rules.

src/components/organisms/caseStudies/CaseStudyFooter.jsx
kesslern: All fixed.

PR #471, opened 2020-03-05
feat(site): Add mission statement

kesslern:
Close #486

image

I noticed at certain sizes the text overlaps the brighter area of the background and the contrast isn't the best, but I'm not sure what the best way to fix it would be or how much it really matters.

image

### DIFF
--- a/src/components/organisms/Hero.jsx
+++ b/src/components/organisms/Hero.jsx
@@ -138,7 +138,7 @@ function Hero() {
         <HeroHeading>
           <BaseTwoLogoText id="hero-logo" />
         </HeroHeading>
-        <HeroTagline> Our mission is to solve problems that matter. We believe that best outcomes start on a solid foundation: open, honest, and equitable partnerships. Together we make exceptional the standard. </HeroTagline>
+        <HeroTagline>Our mission is to solve problems that matter. We believe that best outcomes start on a solid foundation: open, honest, and equitable partnerships. Together we make exceptional the standard.</HeroTagline>
         <ScrollLink target={scroll.target} title={scroll.title}>
           <SatelliteIndicator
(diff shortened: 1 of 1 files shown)


src/components/organisms/Hero.jsx
```diff
@@ -137,7 +138,7 @@ function Hero() {
         <HeroHeading>
           <BaseTwoLogoText id="hero-logo" />
         </HeroHeading>
-        <HeroTagline> We turn ideas into beautiful software </HeroTagline>
+        <HeroTagline> Our mission is to solve problems that matter. We believe that best outcomes start on a solid foundation: open, honest, and equitable partnerships. Together we make exceptional the standard. </HeroTagline>
```
mewdriller: 🤔 Weird that there was previously leading/trailing space in this node. I _think_ that should be safe to remove.

```suggestion
        <HeroTagline>Our mission is to solve problems that matter. We believe that best outcomes start on a solid foundation: open, honest, and equitable partnerships. Together we make exceptional the standard.</HeroTagline>
```

src/components/organisms/Hero.jsx
```diff
@@ -80,9 +80,10 @@ const HeroTagline = styled(P)`
   font-size: 1.15em;
   font-style: italic;
   font-weight: 100;
-  margin-top: 0;
-  margin-bottom: 0.5em;
+  margin: 0 auto 0.5em;
```
mewdriller: 🤔 It looks like something similar to `2em` feels more comfortable for top/bottom margin. But maybe others have thoughts on the aesthetic side of things.

```suggestion
  margin: 2em auto;
```

kesslern: Looks good to me, I'll make the change unless anyone objects.

PR #473, opened 2020-03-10
Browserlist de-duplication & lint improvements

### LINKED ISSUE #207: Remove duplication in browserslist
Currently there is a duplicated list of target browsers in both `.stylelintrc` and `.babelrc`. This can be avoided using the [browserslist](https://github.com/ai/browserslist) field in `package.json` or a separate `browserslist` file.

kesslern:
Close #207

The supported browser list was duplicated in `.babelrc` and `.stylelintrc`. This was moved to `.browserslistrc` to be picked up by all packages supporting browserlist.

Once that was done, I wanted to do some cleanup on the current JS and CSS linting errors, went down a bit of a rabbit hole of linting improvements. I fixed a core-js version warning that appeared during build, and fixed peer dependency warnings by updating React and some eslint rules.  The updated rules required another round of eslint fixes. 

All of the new eslint errors were fixed in the last commit in this PR except for a bunch of dependency cycle errors. I think those would best be tackled in a separate card.

Since this involved a React upgrade, I did a quick test in:
- [x] Safari (Mac/iPad/iPhone)
- [x] Chrome (Linux/Windows/Android)
- [x] Firefox (Windows)
- [x] IE11

Some linting issues fixed include:
- Reordering CSS properties
- Changing two components to functional components
- Moving `propTypes` and `defaultProps` outside of classes (properties instead of using `static`)
- Reordering imports
- Changing imports to use the shortest possible relative path
- Destructure values from `this.state` and `this.props` instead of doing `this.props.foo`


src/components/organisms/ContactUs.jsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import styled from 'styled-components';
-import { ContactForm, ContactHeader, ContactInformation } from '../organisms';
+import { ContactForm, ContactHeader, ContactInformation } from ".";
```
mewdriller: ♻️ Same comment RE dependency cycles.

In this case I think a one-import-per-peer solution is required?

src/components/organisms/jobs/common.jsx
```diff
@@ -61,7 +61,8 @@ const LI = styled.li`
 
 const P = styled.p`
   ${text};
-  &:li {
+
+  & li {
```
mewdriller: 🎉 Good catch.

src/templates/caseStudy.jsx
```diff
@@ -1,25 +1,17 @@
+import { MDXProvider } from '@mdx-js/react';
 import { graphql } from 'gatsby';
+import { MDXRenderer } from 'gatsby-plugin-mdx';
 import { em, rem } from 'polished';
 import PropTypes from 'prop-types';
 import React from 'react';
 import { mapProps } from 'recompose';
 import styled, { ThemeProvider } from 'styled-components';
-import { MDXRenderer } from 'gatsby-plugin-mdx';
+import { CaseStudyAside, CaseStudyFooter, ClientHeader, GlobalNavigation, Main, Section } from '../components';
```
mewdriller: ⚠️ This line is over 100 characters. Did `eslint` / `prettier` not catch this and/or correct it?

src/templates/caseStudy.jsx
```diff
@@ -57,57 +49,6 @@ const CaseStudyContent = styled.article`
...
-
-const PostImg = styled.img`
-  margin: ${em('18px', '18px')} 0;
-
-  ${mediaQuery.small`
-    margin: ${em('21px', '21px')} 0;
-  `};
-`;
-
-const renderMarkdown = raw =>
```
mewdriller: 🎉 Good catch on the dead code elimination here.

.eslintrc
```diff
@@ -14,6 +14,8 @@
         "components": ["Link"],
         "specialLink": ["to"]
       }
-    ]
+    ],
+    "react/jsx-one-expression-per-line": "off",
+    "react/jsx-props-no-spreading": "off"
```
mewdriller: 🤔 Interesting that Airbnb is now forbidding JSX spread. I'm not opposed to that in non-TypeScript projects; I've been burned a few times trying to trace what props are actually used in nested-spread scenarios.

src/components/atoms/form/TextField.jsx
```diff
@@ -11,49 +11,38 @@ const stateUpdateFromValue = ({ value }) => ({ value, isDirty: !!value });
 const floatLabel = ({ type }) => type === 'file';
 
 class TextField extends React.Component {
```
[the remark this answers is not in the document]
kesslern: Created https://github.com/b2io/base2.io/issues/474

src/components/atoms/form/TextField.jsx
```diff
@@ -79,6 +68,25 @@ class TextField extends React.Component {
...
+  id: PropTypes.string,
+  multiline: PropTypes.bool,
+  onBlur: PropTypes.func,
+  onChange: PropTypes.func,
+  onFocus: PropTypes.func,
+  value: PropTypes.string,
+};
+
 function TextAreaField(props) {
   return <TextField {...props} input={Textarea} />;
```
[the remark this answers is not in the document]
kesslern: https://github.com/b2io/base2.io/issues/475

src/components/molecules/BlogHeader.jsx
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import styled from 'styled-components';
 import { cssEvery, mediaQuery, themed } from '../../util/style';
-import { Header, Time } from '../../components';
+import { Header, Time } from "..";
```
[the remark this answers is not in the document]
kesslern: I'll see what I can do about this. It seems to be working fine, and the updated linting rules are pointing out a bunch of dependency cycle errors. I didn't try to fix them yet, so I'll take a look.

src/components/organisms/jobs/common.jsx
kesslern: This was actually caught by the updated stylelint rules.

src/templates/caseStudy.jsx
kesslern: I didn't realize it wasn't part of the build. I installed the vscode prettier extension and set it to format on save. I'll go ahead and re-format this file too.

src/components/organisms/ContactUs.jsx
kesslern: This is fixed by changing the import to `./contact`

src/components/molecules/BlogHeader.jsx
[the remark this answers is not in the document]
kesslern: Thanks for the pointer, I got all the dependency cycle errors fixed. There are now 0 eslint errors in the project :man_dancing:

Issue #316, opened 2018-05-02
Code blocks syntax highlighting


reeddunkle:
As a User, I'd like code blocks in blog posts to optionally have syntax highlighting, such as:

```js
const isJavaScriptHighlighting = true;
function add(a, b) {
  return a + b;
}
```

Related: #322, #358

mpcondo:
closing this issue because it was a part of PR #476

PR #476, opened 2020-03-11
feat(site): Add syntax highlighting for blog posts

### LINKED ISSUE #452: Blog code blocks
The code blocks in blogs, such as the git-configured post, are not displaying

kesslern:
Close #452 and #316 

I went with a dark theme for syntax highlighting to match the rest of the app. I could also see using a lighter theme since the blog posts are on a white background. Let me know what you think!

* Create a `HighlightedPre` atom using `prism-react-renderer`
* Add syntax highlighting to blog posts using `HighlightedPre` to render `pre` tags via `MDXProvider` APIs
* Add support for ruby highlighting, which is not included with `prism-react-renderer` by default
* Add language to existing blog posts
* Syntax highlighting is supported in both SSR and CSR
* Fix width of blog post content on wide screens (it was 200px narrower than the Figma designs call for)

Example of highlighted Ruby code:
Highlighted ruby code

Tested on:
- [x] Safari (iPhone/Mac)
- [x] IE11
- [x] Chrome (Linux/Android/Windows)
- [x] Firefox (Windows)


src/templates/caseStudy.jsx
```diff
@@ -1,11 +1,11 @@
-import { MDXProvider } from '@mdx-js/react';
 import { graphql } from 'gatsby';
 import { MDXRenderer } from 'gatsby-plugin-mdx';
 import { em, rem } from 'polished';
 import PropTypes from 'prop-types';
 import React from 'react';
 import { mapProps } from 'recompose';
 import styled, { ThemeProvider } from 'styled-components';
+import { MDXProvider } from '@mdx-js/react';
```
mewdriller: ⚠️ It looks like this got pulled out of alpha-order for imports unnecessarily.

src/templates/post.jsx
```diff
@@ -31,27 +31,24 @@ const PostContent = styled(Section)`
...
-    </ThemeProvider>
+    <MDXProvider components={defaultComponentMap}>
+      <ThemeProvider theme={lightTheme}>
+        <Main>
+          <GlobalNavigation />
+          <BlogHeader
+            author={author}
+            img="/img/transmission-constellation.png"
+            imgAlt="Satellite broadcasting into space"
+            date={date}
```
mewdriller: 🤔 Alpha-order on props is off but I'm guessing it's probably not setup in the `eslint` rules. Maybe something to do in another task?

gatsby-browser.js
```diff
@@ -3,6 +3,7 @@ import wrapWithGlobalStyles from './wrap-with-global-styles';
 export const onClientEntry = () => {
   // Polyfills:
   require('whatwg-fetch');
+  require('./src/util/prism-languages');
```
mewdriller: ⚠️ If I remember correctly, `onClientEntry` only runs in the browser context. It looks like `src/util/prism-languages` includes some SSR-safe code that might be extraneous. 

It also might be better to isolate the weight of the `prism` imports to only those pages that would use it. That's _probably_ only blog posts? But maybe we're using it elsewhere.

src/components/atoms/HighlightedPre.jsx
```diff
@@ -0,0 +1,39 @@
...
+
+  return (
+    <Highlight
+      {...defaultProps}
+      code={props.children.props.children}
+      language={
+        matches && matches.groups && matches.groups.lang
+          ? matches.groups.lang
+          : ''
+      }
```
mewdriller: 🤔 Do we have optional chaining and nullish coalescing in this project? If so this can be simplified to:

```suggestion
      language={matches?.groups?.lang ?? ''}
```

src/util/templates.jsx
```diff
@@ -77,7 +77,7 @@ export const defaultComponentMap = {
   li: LI,
   ol: OL,
   p: P,
-  pre: Pre,
+  pre: HighlightedPre,
```
mewdriller: ⚠️ I _think_ there's a to detect the language off the triple-backtick syntax; at least there's a way in most markdown processors. It may make sense to defer the cost of the `prisma` import + load until we're sure that a particular `pre` actually uses syntax highlighting.

Something like [`loadable-components`](https://github.com/gregberge/loadable-components) could help w/ deferring the import cost.

src/components/atoms/HighlightedPre.jsx
kesslern: Looks like we do, nice!

src/templates/caseStudy.jsx
kesslern: Fixed

src/templates/post.jsx
kesslern: This is auto-fixable once the eslint rule is added, and there's 55 instances of out-of-order props across many files, so I'll add that rule and fix all the instances in another PR.

gatsby-browser.js
kesslern: I did a little more experimentation with this, and since the syntax highlighting is able to be handled server-side, I'm able to remove it from the browser context entirely. I did need to _add_ `prism-languages` to the SSR build by importing it in `gatsby-ssr.js`. Before I made that change, Ruby highlighting would be done client-side. Now, all syntax highlighting works even if the user has JS completely disabled.

src/util/templates.jsx
kesslern: I might be misunderstanding something about how Gatsby works, but I think by moving all the highlighting to the SSR we avoid having to ever push prism out to the browser. The component tree rendered by Gatsby when the app loads just contains the `Pre` component produced by the `HighlightedPre` and prism, with all the styles included inline.

mewdriller: That's even better! 🎉

PR #484, opened 2020-03-20
fix(site): Fix "View Our Work" button

### LINKED ISSUE #477: Bug: "View Our Work" link unclickable
The "View our work" link seems to be cut off and is unclickable.
Screen Shot 2020-03-11 at 3 29 42 PM

kesslern:
Close #477

The wrapper in the Clients section was intended to have a padding at the bottom where a linear gradient would sit. The padding was getting overwritten by the standard container styles, and then linear gradient was in the same space.

To fix this, I changed the common container styles to only set the left and right padding, which is I think what it was intended for.

### DIFF
--- a/src/util/style.js
+++ b/src/util/style.js
@@ -31,14 +31,17 @@ const mediaQuery = {
 };
 
 const containerSizing = css`
-  padding: 0 5%;
+  padding-left: 5%;
+  padding-right: 5%;
 
   ${mediaQuery.medium`
-    padding: 0 2em;
+    padding-left: 2em;
+    padding-right: 2em;
   `};
 
   ${mediaQuery.large`
-    padding: 0 100px;
+    padding-left: 100px;
+    padding-right: 100px;
   `};
 `;
 


src/util/style.js
```diff
@@ -31,14 +31,17 @@ const mediaQuery = {
 };
 
 const containerSizing = css`
-  padding: 0 5%;
+  padding-left: 5%;
+  padding-right: 5%;
```
mewdriller: 🤔 Interesting. This changes my perspective on the always use the shortcut rules I'm used to. Does something like `inherit 5%` keep the previous rule's value for top/bottom?

kesslern: I sure do wish that'd work, but apparently `inherit` is only supported as a single property. You can combine the shorthand and then override part of it, because only the last rule is used:

```js
const containerSizing = css`
  padding: 5% 0;
  padding-left: inherit;
  padding-right: inherit;
`

Could be useful in some cases, but in this case I don't think it's any better.

PR #485, opened 2020-03-23
fix(site): Blog Post Feature Images

### LINKED ISSUE #457: Blog post feature images
Display images in blog post list (similar to the case study list images) by utilizing the meta data `image` when it exists. 

Note: This logic had been in the case-studies-feature-branch prior to master being merged down into it. The feature was lost due to merge conflicts.

kesslern:
Close #457 

Add feature image support (back) to the blog list.

[Uploading BlogPostList.png…]()

### DIFF
--- a/src/pages/blog/index.jsx
+++ b/src/pages/blog/index.jsx
@@ -76,6 +76,7 @@ function mapPropsToProps({ data }) {
       author: authorIdToName[frontmatter.author],
       date: frontmatter.date,
       id: frontmatter.path,
+      featureImg: frontmatter.image || undefined,
       path: frontmatter.path,
       title: node.frontmatter.title,
     };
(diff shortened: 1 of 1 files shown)


src/pages/blog/index.jsx
```diff
@@ -76,6 +76,7 @@ function mapPropsToProps({ data }) {
       author: authorIdToName[frontmatter.author],
       date: frontmatter.date,
       id: frontmatter.path,
+      featureImg: frontmatter.image || undefined,
```
mewdriller: 🤔 Does `frontmatter.image` come through as `''` or something when it's not set in the underlying `md` file?

kesslern: It comes through as `null` when it's not defined, which overrides the `defaultProps`. I've no idea why it comes through as `null` instead of `undefined`.

PR #488, opened 2020-05-13
474 convert textfield

### LINKED ISSUE #474: Convert TextField to an SFC
The `TextField` component uses the old `extends React.component` definition and makes use of `  UNSAFE_componentWillReceiveProps()`. This component should be updated to a functional component, and can make use of the Hooks API.

mpcondo:
Replaced UNSAFE_componentWillReceiveProps with componentWillReceiveProps  in order to replace the deprecated lifecycle method with a new lifecycle method compatible with React 17.

closes #474


kesslern:
The fix looks good to me... but is this component even used anywhere?

PR #490, opened 2020-05-18
456 technology tooltip

### LINKED ISSUE #456: Technology tooltips
Display the name of the technology as a tooltip with the technology icon within case studies is hovered over.

mpcondo:
Closes #456 

-Added react tooltip to the project
-Added the tooltip to CaseStudyAside.jsx

### DIFF
--- a/src/components/organisms/caseStudies/CaseStudyAside.jsx
+++ b/src/components/organisms/caseStudies/CaseStudyAside.jsx
@@ -79,7 +79,7 @@ function CaseStudyAside({ client, highlights, link, project, technologies }) {
 
       <div>
         {mapToTech}
-        <ReactTooltip />
+        <ReactTooltip effect="solid" />
       </div>
     </CaseStudyAsideWrapper>
   );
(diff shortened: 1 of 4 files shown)


src/components/organisms/caseStudies/CaseStudyAside.jsx
```diff
@@ -62,13 +78,8 @@ function CaseStudyAside({ client, highlights, link, project, technologies }) {
...
       <div>
-        {technologies.map(tech => (
-          <CaseStudyTechnologyLogo
-            {...tech.image}
-            alt={tech.name}
-            key={tech.name}
-          />
-        ))}
+        {mapToTech}
+        <ReactTooltip />
```
mewdriller, line 82: ⚠️ I would expect the tooltip to anchor to the element rather than the mouse. The user experience is a bit worse in this mode as the tooltip can end up obscuring portions of the logo depending on where your cursor is.

image

image

src/components/organisms/caseStudies/CaseStudyAside.jsx
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 import styled from 'styled-components';
 import PropTypes from 'prop-types';
 import { em, rem } from 'polished';
+import ReactTooltip from 'react-tooltip';
```
mewdriller, line 5: ⚠️ To be consistent, I think we should probably use the same tooltip behavior on the grid on the root index as well.

image

PR #513, opened 2021-03-11
story: CTA

tborres:
Closes [B2IO-3](https://b2io.atlassian.net/browse/B2IO-3)

Simple CTA with no animation


emotion.d.ts
```diff
@@ -0,0 +1,8 @@
+import '@emotion/react';
+
+import { ThemeType } from './src/theme';
+
+declare module '@emotion/react' {
```
tborres, line 5: Resolve `theme.color` typing errors when used in emotion styling

PR #517, opened 2021-03-18
chore: breakpoint setup

tborres:
This PR defines breakpoint sizes (based on [bootstrap](https://getbootstrap.com/docs/5.0/layout/breakpoints/)), and introduces utility constants for quick media queries.

Closes [B2IO-4](https://b2io.atlassian.net/browse/B2IO-4?atlOrigin=eyJpIjoiNzk1NjFlYTAxNmY0NDdhZWEwN2FjZTIxZWJmODlhNjEiLCJwIjoiaiJ9) #done


src/breakpoints.ts
```diff
@@ -0,0 +1,17 @@
+export const bp = {
+  xs: 0,
+  sm: 576,
+  md: 768,
+  lg: 992,
+  xl: 1200,
+  xxl: 1400,
```
sludemann, line 7: These don't seem to match the description on the card.  Was that intentional?

sludemann (approved):
Changes make sense, just one requirements question.

src/breakpoints.ts
tborres: Yes(ish). The description on the card was a quick eyeballing of what sizes we had in the BFR designs. This list is based on the breakpoints that bootstrap uses.

tborres: I've updated the card descript to be even more vague but less contradictory.

PR #521, opened 2021-03-22
task: header

tborres:
Closes [B2-7](https://b2io.atlassian.net/browse/B2-7) #comment
Closes [B2-8](https://b2io.atlassian.net/browse/B2-8) #comment

- move nav out of `Layout` and into a `Header` component
- display mobile menu icon on small screen
- responsive header layout
- move Logo to a `LogoIcon` within an `icons` directory
- shift `LogoProps` to be a shared `SvgProps` type
- update to typescript 4 for `react-use` library
- add mobile menu and animations


src/components/Header.tsx
```diff
@@ -0,0 +1,135 @@
...
+  width: 100%;
+
+  ${atMd} {
+    padding: 36px 40px 87px;
+  }
+
+  ${atLg} {
+    padding: 32px 80px 87px;
+  }
+`;
```
tborres, line 36: I don't love how this jumps around with the screen size changing.

src/components/Header.tsx
```diff
@@ -0,0 +1,135 @@
+import Link from 'next/link';
+import React, { FC } from 'react';
+import { LogoWithName } from './';
+import { atLg, atMd, atSm } from '~/breakpoints';
+import styled from '@emotion/styled';
+import { css } from '@emotion/react';
+import { MenuIcon } from './icons';
+import { useRouter } from 'next/router';
+
+const Container = styled.header`
```
chriswhawkins, line 10: Consider moving the gradient effect to a pseudo-element, or using some other (theme driven? 🤔) mechanism for max-width on the contents.

image

README.md
```diff
@@ -89,9 +89,9 @@ Starts storybook on port 6006. Storybook is a demonstration home for the library
     <summary>If a new card is needed</summary>   
     If you need to create a new card then click the `Create` button in Jira's header and select the correct card type
 
-     - **story:** A user-focused feature written as a user story.
```
tborres, line 92: 💬 Just misc auto formatting corrections in this file.

package.json
```diff
@@ -16,8 +16,10 @@
     "@emotion/styled": "^11.0.0",
     "framer-motion": "^3.3.0",
     "next": "latest",
+    "normalize.css": "^8.0.1",
```
tborres, line 19: 💬 normalize and reset css to remove browser defaults like 8px padding on body.

package.json
```diff
@@ -16,8 +16,10 @@
...
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
+    "react-use": "^17.2.1",
```
tborres, line 22: 💬 Used for `useLockBodyScroll` in the mobile menu.

package.json
```diff
@@ -47,7 +49,7 @@
     "jest-watch-typeahead": "^0.6.1",
     "lint-staged": "^10.5.3",
     "prettier": "^2.2.1",
-    "typescript": "^3.7.4"
+    "typescript": "^4.2.3"
```
tborres, line 52: 💬 Typescript update was needed to avoid type errors with the `react-use` library.

src/components/Header.tsx
tborres: The clipping problem is no longer an issue now that the css has been normalized and the gradient has been moved. I do think that we'll want to store the gradient and/or max-width in the theme or a global file, but I don't feel comfortable enough with the design files at this point to find a good home for them.

src/components/Layout.tsx
```diff
@@ -43,47 +43,16 @@ export const Layout: FC<LayoutProps> = ({ children, title = 'Base Two' }) => {
...
-              <Link href="/contact">
-                <a>Contact</a>
-              </Link>
-            </li>
-          </ul>
-        </nav>
-      </header>
+      <Header
+        css={css`
+          max-width: 1600px;
```
tborres, line 48: 💬 This max-width and margin-top will need to be reconsidered at a global level, but I think it's ok to move forward as is.

src/components/MobileMenu.tsx
```diff
@@ -0,0 +1,118 @@
...
+import Link from 'next/link';
+import React, { FC } from 'react';
+import { useLockBodyScroll, useToggle } from 'react-use';
+import { CTA } from '.';
+import { MenuIcon } from './icons';
+
+const textAnimation = `
+  opacity: 0;
+
+  .visible & {
```
k8hansen, line 11: @chriswhawkins, should we be using Aria hidden/expand attributes instead of straight classes for something like this?

src/components/MobileMenu.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+const Menu = styled.div`
+  background-color: ${({ theme }) => theme.colors.workThumbnailBg};
+  height: 100%;
+  left: 0;
+  position: fixed;
+  top: 0;
+  transform: scale(0, 0);
+  transform-origin: top right;
+  transition: transform 100ms ease-in-out;
```
k8hansen, line 25:
```suggestion
  transition: transform 220ms ease-out, opacity 220ms ease-out;
```
I'm not loving this fix either, but I think it is improved just visually. I would appreciate someone else's opinion on making this animation a little smoother.

k8hansen (commented):
Minor animation suggestions, but my bigger suggestion would be using aria attributes for menu toggling instead of classes.

src/components/MobileMenu.tsx
```diff
@@ -0,0 +1,118 @@
...
+    opacity: 1;
+    transition: opacity 500ms ease 100ms;
+  }
+`;
+
+const Menu = styled.div`
+  background-color: ${({ theme }) => theme.colors.workThumbnailBg};
+  height: 100%;
+  left: 0;
+  position: fixed;
```
k8hansen, line 21:
```suggestion
  opacity: 0;
```
begin opacity at zero before updated transition below.

src/components/MobileMenu.tsx
```diff
@@ -0,0 +1,118 @@
...
+  top: 0;
+  transform: scale(0, 0);
+  transform-origin: top right;
+  transition: transform 100ms ease-in-out;
+  width: 100%;
+  z-index: -1;
+
```
k8hansen, line 28:
```suggestion
opacity: 0;
```
I believe you need to add this as a baseline before applying transition below

src/components/MobileMenu.tsx
```diff
@@ -0,0 +1,118 @@
...
+  &.visible {
+    transform: scale(1, 1);
+  }
```
k8hansen, line 31:
```suggestion
  }
  opacity: 1;
```

k8hansen, line 31:
```suggestion
opacity: 1;
    }
```

src/components/MobileMenu.tsx
tborres: I think that it would depend on what we want to do with the transition. Currently the transition is just a scale up from 0 (the text is a fade in, which is separate). Adding this opacity line causes the menu to never show (unless we also want to transition the opacity up to 1).

src/pages/_app.tsx
```diff
@@ -25,6 +25,8 @@ const CustomApp: FC<AppProps> = ({ Component, pageProps }) => {
               * {
                 box-sizing: border-box;
                 font-family: Roobert;
+                margin: 0;
+                padding: 0;
```
FrankyMartz, line 29: Recommend against globally defaulting HTMLElements to a zero padding. Have personally found least resistance using the Normalize.CSS library instead of outright starting from zero, as CSS resets require more intensive testing on far more clients.

src/components/Header.tsx
```diff
@@ -0,0 +1,143 @@
...
+`;
+
+const Content = styled.div`
+  align-items: center;
+  display: flex;
+  justify-content: space-between;
+  margin: 0 auto;
+  padding: 24px 24px 40px;
+
+  ${atMd} {
```
FrankyMartz, line 29: Coolio, 👍🏽 . We will have to adjust menu item spacing and sizes but definitely a solid kick-off.

FrankyMartz (changes requested):
It looks like the Mobile menu disappears, because the header width grows beyond the widget boundary.
<details>
<summary>Click here to view</summary>
Screen Shot 2021-03-26 at 10 24 27 PM
</details>

In terms of menu animation, I suggest one of two patterns (I'll expand on second option further down). For right now though, we should pick a single axis (x,y) and "Slide-in" on that axis. So for instance, menu button click may slide the menu backdrop (top -> bottom) or (right -> left).
 The reason for this is both simplicity and perception. A moving edge corner-to-corner reads uncomfortable (sharp edges), and we want to use a motion the user is familiar with like blinds, door, etc because this "feels" inviting. The other note, is that even if a corner-to-corner animation may have the same duration as a single axis motion. The perceived distance to travel in corner-to-corner feels longer. We do want to transition the menu in but we don't want the user to feel as though they have to wait for it (even if they actually aren't).
 
On other concept: I've been playing around with the idea smart glass. There's this special glass that exists that is transparent by default but when the space requires privacy, isolation the glass will blur-in as frosted.  I've been playing with this idea, envisioning this for our menu as it could animate relative quickly and lack of distance to travel would reduce perceived duration.

src/pages/_app.tsx
k8hansen: Interesting!

package.json
FrankyMartz, line 19: We should really include NodeJS version.
```json
  "engines": {
    "node": "^15.12.0",
    "npm": "^7.6.3"
  }
```
Have noticed with last few clean `npm install`'s that the  `package-lock.json` file is updating.  This should not happen when a `package-lock.json` is committed to the repository. The only time that happens is when there is significant difference in NodeJS versions utilized between developers or a npm package was removed/added.

src/components/Layout.tsx
FrankyMartz: When it comes to adaptive layouts, you want to avoid introducing axis limitations in children (explicit width/height) and instead use relative portions — i.e. percentages where possible. The concept being each child element should only concern itself with accommodating itself within the space allotted by its parent, and should not concern itself with the explicit dimensions/limitations of that parent.

 <details>
 <summary>Click here to view example</summary>
Screen Shot 2021-03-30 at 10 32 34 PM
</details>
  
  In order to prevent un-viewable overflow on the page you would use closest element near `body` or in rare cases `body` to establish `max-width/min-width` limitation.

src/components/Header.tsx
```diff
@@ -0,0 +1,143 @@
...
+const Container = styled.header`
+  background-image: linear-gradient(
+    to bottom,
+    rgba(4, 0, 28, 0.84) 35%,
+    rgba(6, 2, 27, 0)
+  );
+  height: 146px;
+  position: fixed;
+  top: 0;
+  width: 100vw;
```
FrankyMartz, line 19: This element should span the width of the parent, ideally `main` in a `body -> main` configuration. Using an explicit width here is likely to introduce issues down the line as `100vw` may not be possible should it's ancestors introduce padding/margins.

FrankyMartz, line 19: This is currently not achieving full width span in tablet views due to `max-width` implementation in child element `<Content>`. Also, recommend using `100%` vs `100vw`; since in some situations `vw` includes scrollbars in width calculation resulting in unintended horizontal scroll.

FrankyMartz (changes requested):
The doesn't span full width on smaller devices. This should be addressed.
<details>
<summary>Click here to view</summary>
Screen Shot 2021-03-30 at 11 21 55 PM
</details>

The sidebar animation is much better 👍🏽 !
We may eventually want to include an animation to hide menu instead of simply having it disappear, but that can be tweaked later.

package.json
tborres: The primary driver of the lock file being updated recently is the addition of the normalize.css package. Let's chat about locking the node version and which version to lock it to. I can see benefits of locking it tightly to a version to reduce potential differences between machines, but that is at the cost of forcing contributors on the team to switch their node version often while dipping into the project from their other projects.

src/components/Header.tsx
tborres: 🙁 I had switch from `100%` to `100vw` to try to resolve the "header width grows beyond the widget boundary" issue, but I think the header growing is more of a consequence of the crazy font sizes that are currently in the main body of the site rather than an issue with the header. For instance, when I greatly increase the font sizes on the Bootstrap site it has the same type of header sizing issue.

I'll switch it back to 100% and then address the sizing issues at a future time if they continue when we have real content. However, I'm open to specific recommended solutions.

PR #526, opened 2021-04-02
feat(Container): Organize Theme, Container, Breakpoints

FrankyMartz:
## Description

- refactor `theme.ts` into directory
- add `spacing` to theme
- update `Breakpoints` with more utilities
- add `Container` component for boundary restriction

### Note

- I've opted for use of a `<Container />` component instead of baking in the width limitation into `body->first-child`. This decision was made in consideration of ease of adaption. Should a section require it's assets to break out of content area it's easier to target content-only sections. In terms of implementation, the `<Container/>` component would be used to wrap content which should adhere to max-width restriction from designs, and should ONLY be nested under parents who do NOT enforce any x-axis restrictions that affect width (width, padding-x, margin )

#### 2021-04-02

- As a means to control overflow from preventing vertical scrollbars, I've move the `Layout` children into a div element that hides overflow, and then post Motion element nested  the children in `Container` to manage `padding`s. The reason I've relocated overflow hidden to parent is because we want the viewable area not to hide overflow as Container is only intended to contain content to screen center and not obscure assets that break it's bounds.
- Pages are responsive as breakpoints only affect paddings until a `width` of `1440px` is reached; at which point the `max-width` limitation is enacted per designs. 

## Related Stories

Closes [b2-62: Container for Resize Foundation](https://b2io.atlassian.net/browse/B2-62)


src/theme/index.ts
```diff
@@ -0,0 +1,18 @@
...
+import { spacing } from './spacing';
+import { textVariants } from './typography';
+
+const theme = {
+  colors,
+  spacing,
+  textVariants,
+};
+
+export type ThemeColor = keyof typeof colors;
```
tborres, line 15: 💬 This same export line exists in `colors.ts`. Should it just be imported from `colors.ts` and re-exported here?

src/theme/index.ts
```diff
@@ -0,0 +1,18 @@
+/* ========================================================================== *
+ * THEME
+ * ========================================================================== */
+
```
tborres, line 4: 💬 Just curious, why add this comment at the top? Is this because it's an index file and index files are hard to understand the purpose of without context?

src/theme/typography.ts
```diff
@@ -1,20 +1,6 @@
...
-  action: palette.coral,
-  background: palette.darkBlue,
-  text: palette.offWhite,
-};
-
-export type ThemeColor = keyof typeof theme.colors;
+import { colors } from './colors';
 
-const textVariants = {
+export const textVariants = {
```
tborres, line 3: As a follow-up task (after this PR) we should update the typography to match the comps from a mobile first perspective. The current styling doesn't seem to match the comps and doesn't adjust based on screen size.

src/pages/index.tsx
```diff
@@ -1,31 +1,39 @@
...
-        </Text>
-      </Heading>
-      <Heading as="h2" variant="h2">
-        Fueled by{' '}
-        <Text as="span" variant="hero">
-          curiosity.
-        </Text>
-      </Heading>
-      <CTA href="/404">404</CTA>
+      {/* NOTE: Container intended for in section component implementation */}
```
tborres, line 13: I'm confused by the purpose of this comment. Is this meant to be an explanation as to why it wasn't just added to the `Layout` component?

tborres (commented):
Good changes overall. I just have a few questions and callouts that I'd like to discuss before merging.

- [x] Since we don't have the node versions locked in I think that we should only commit the `package-lock.json` in the case of a change in the `package.json` otherwise we will create more change waves and collisions than necessary. I'm open to discussion on this (and honestly I don't really care that much)
- [ ] the current Container implementation is resulting in nested scrollbars. How should these be addressed (see below)
- [x] Why build the Container to be used at a component level rather than incorporated into the Layout?
- [ ] What are your thoughts moving forward (after this PR) for the typography theming (e.g. `px` vs `em` and responsiveness)

image

src/theme/index.ts
FrankyMartz: Sorry. Add these when I'm juggling buffers.
🔥 Comment Header removed

src/theme/index.ts
FrankyMartz: Oops, leftover from when I broke out `Theme.ts` into directory of modules.
Refactored to export from files instead.

src/pages/index.tsx
FrankyMartz: I wanted to relay that the `Container` would be used in each page section as required. I've since removed that since I've refactored Layout to utilize `Container` to wrap all the contents of all pages through the `Layout` component

src/pages/index.tsx
```diff
@@ -21,7 +22,14 @@ const HomePage: NextPage = () => {
...
         Fueled by{' '}
-        <Text as="span" variant="hero">
+        <Text
+          as="span"
+          css={css`
+            position: relative;
+            left: 50%;
+          `}
+          variant="hero"
+        >
```
tborres, line 32: ❓ Was this added for testing purposes?

```suggestion
        <Text as="span" variant="hero">
```

Note: the `css` import would need to be removed too.

src/components/Layout.tsx
```diff
@@ -4,7 +4,12 @@ import Head from 'next/head';
 import Link from 'next/link';
 import React, { FC } from 'react';
 
-import { Header } from './';
+import { Container, Header } from './';
+import styled from '@emotion/styled';
+
+const Root = styled.div`
+  overflow: hidden;
```
tborres, line 11: 💬 This prevents horizontal page scrolling, which I'm not sure is our goal. In my tests the `Root` doesn't seem to be necessary to prevent the nested vertical scrollbars.

tborres (approved):
I added a few comments for consideration / clean-up, but this is ready to merge at your discretion.

src/pages/index.tsx
FrankyMartz: Yes, I modified to verify content area was expanding/collapsing correctly. Thought it should be fine since all of that content there is only placeholder for actual page.

src/components/Layout.tsx
FrankyMartz: The multiple vertical scrollbar bug was resolved when I refactored out each individual `Container` use for a single `Container` implementation at the `Layout` component level.

Refactoring into a single `Container` resulted in an issue where the (view width)`vw` was dynamically growing to accommodate its content. This resulted in the menu bar/hamburger being shoved to the right and off the view; it also resulted in a horizontal scrollbar for `<main/>`. This bug didn't show up for all browsers, noted it functioned as expected in FF but was severely broken in Chrome on mobile mode.

Please NOTE: While, I did test this out in FF, Safari, Chrome, MS Edge; I can't guarantee 💯 bug proof as that level of confidence can not be achieved with out all of the pages contents. CSS bugs are always circumstantial to very specific combinations; as any child element is capable of modifying layout and affecting its responsive parent's dimension. This implementation should get us started and I'm here to help when issues arise.

PR #528, opened 2021-04-02
chore: set the correct font-feature-settings

tborres:


tborres:
FYI: @k8hansen @FrankyMartz this corrects the font to match the one used in the design files.

PR #531, opened 2021-04-06
chore: remove storybook

tborres:
Since the emotion `css` doesn't behave the same way in storybook as it
does in the site the decision was made to remove storybook all together.

Closes [B2-65](https://b2io.atlassian.net/browse/B2-65)


tborres:
#532 resolves the `css` + `storybook` issue so that we don't need to burn the whole thing to the ground.

PR #533, opened 2021-04-07
task: create and use typography theming

tborres:
- update typography variables to match the zeplin styleguide
- update the font variants
- update the app to use `rem`s instead of `px` and variables when possible

Closes [B2-66](https://b2io.atlassian.net/browse/B2-66)

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -12,6 +12,7 @@ module.exports = {
         ...config.resolve,
         alias: {
           ...config.resolve.alias,
+          '~': toPath('src/'),
           '@emotion/styled': toPath('node_modules/@emotion/styled'),
         },
       },
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -23,7 +23,7 @@ export const decorators = [
           * {
             box-sizing: border-box;
             font-family: Roobert;
-            font-feature-settings: 'ss02';
+            font-feature-settings: 'ss02', 'ss05';
           }
         `}
       />
--- a/src/components/CTA.tsx
+++ b/src/components/CTA.tsx
@@ -8,19 +8,16 @@ export type CTAProps = {
 };
 
 const A = styled.a(({ theme }) => ({
-  color: theme.colors.text,
+  ...theme.textVariants.CTA,
   cursor: 'pointer',
   display: 'inline-block',
-  fontSize: 22,
-  fontWeight: 900,
-  lineHeight: '22px',
   textDecoration: 'none',
   '::after': {
     backgroundColor: theme.colors.action,
     content: '""',
     display: 'block',
-    height: 2,
-    marginTop: 8,
+    height: '0.125rem',
+    marginTop: theme.spacing.xxs,
     width: '100%',
   },
   ':hover': {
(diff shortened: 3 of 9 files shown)


.storybook/main.js
```diff
@@ -12,6 +12,7 @@ module.exports = {
         ...config.resolve,
         alias: {
           ...config.resolve.alias,
+          '~': toPath('src/'),
```
tborres, line 15: 💬 Fixes issue when importing via `~` in files that storybook touches.

.storybook/preview.js
```diff
@@ -23,7 +23,7 @@ export const decorators = [
           * {
             box-sizing: border-box;
             font-family: Roobert;
-            font-feature-settings: 'ss02';
+            font-feature-settings: 'ss02', 'ss05';
```
tborres, line 26: 💬 Fancy quote. I had visually checked and compared each of the variations to the design files.

src/components/CTA.tsx
```diff
@@ -8,19 +8,16 @@ export type CTAProps = {
 };
 
 const A = styled.a(({ theme }) => ({
-  color: theme.colors.text,
+  ...theme.textVariants.CTA,
```
tborres, line 11: 💬 I'm interested to hear if there is a better way to use the `textVariants` than this.

src/components/Header.tsx
```diff
@@ -59,24 +59,35 @@ const Nav = styled.nav`
...
     }
   }
+`;
 
-  a {
-    color: ${(props) => props.theme.colors.offWhite};
-    font-size: 17px;
-    line-height: 28px;
-    text-decoration: none;
+const A = styled.a(({ theme }) => ({
```
tborres, line 72: 💬 I switched to an object-based styled component to be able to spread the textVariant styling. I'm not sure if there is a better / more appropriate way to do this.

src/theme/typography.ts
```diff
@@ -1,57 +1,92 @@
...
 
 export const textVariants = {
   body: {
     color: colors.text,
-    fontSize: '17px',
-    letterSpacing: '0px',
-    lineHeight: '28px',
+    fontSize: '1.063rem',
+    fontVariationSettings: '"wght" 500',
+    letterSpacing: '0.006875rem',
```
tborres, line 9: 💬 This (and similar) letter spacing is per the Zeplin styleguide, but it doesn't seem that it's actually being used in the designs (or the designs simply round it down to 0).

.storybook/preview.js
k8hansen: Wasn't entirely sure how to test this, but I'm seeing a fancy 'g' in storybook and it runs so I consider it a pass.

k8hansen (approved):
:shipit:

.storybook/preview.js
tborres: `ss02` = fancy g
`ss05` = fancy quote.
☝️ technical terms

tborres:
I'm going to merge this in now so that it can start being used, but I'm still interested to hear feedback from anybody who's got some.

PR #534, opened 2021-04-07
chore(storybook): a11y addon

chriswhawkins:
Adds default axe core accessibility checks to storybook component view.

https://storybook.js.org/addons/@storybook/addon-a11y
https://github.com/dequelabs/axe-core


tborres (approved):
Accessibility checks in storybook👍 
image

We may want to consider adding checks to the site itself via Github actions, but it may be a bit early in the dev process for that.

PR #535, opened 2021-04-07
feat(home) static hero

mewdriller:
- Implement static version of the Home page's Hero component
- Add a utility for responsive value clamping
- Upgrade to Webpack 5
- Switch relevant Heading variant values to rem from px
- Refactor related files into the `theme` module
- Fix issue with mobile nav being overlaid by hero w/ `position: relative`

B2-10


src/components/util.ts
```diff
@@ -0,0 +1,20 @@
+export type ScaledFontSizeArgs = {
+  maxValue: number;
+  maxWidth: number;
+  minValue: number;
+  minWidth: number;
+};
+
+export const responsiveViewportRelativeSize = ({
```
tborres, line 8: 💬 Take a look at the updated the typography file. I think that it either make this unnecessary or we should be leveraging this in the typography file.

tborres (commented):
Added a handful of comments - mainly resulting from the typography theme changes that were happening in tandem. Take a look and let's chat about the best direction to move forward.

src/components/index.ts
```diff
@@ -8,3 +8,5 @@ export * from './Text';
 export * from './MobileMenu';
 export * from './Header';
 export * from './Layout';
+
+export * from './util';
```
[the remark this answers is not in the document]
mewdriller: Yeah, I think that's a better spot. I'd like to circle-back on the alpha-order breakage too. That seems like something else must be off.

src/pages/index.tsx
```diff
@@ -1,24 +1,82 @@
...
 
 const HomePage: NextPage = () => {
   return (
     <Layout>
-      <Heading as="h1" variant="hero">
-        Hero
-      </Heading>
-      <Heading as="h1" variant="h1">
-        H1
+      <div
```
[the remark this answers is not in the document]
mewdriller: 🤔 I went back and forth on that. I think hoisting it to a shared folder is too much, but I think the semantic extraction within the same file probably makes sense. I'll try it that way and you can see what you think.

src/pages/index.tsx
```diff
@@ -1,24 +1,82 @@
...
+        <Heading
+          as="h1"
+          css={css`
+            font-size: ${responsiveViewportRelativeSize({
+              maxValue: 6.25,
+              maxWidth: bp.lg / 16,
+              minValue: 3.75,
+              minWidth: (bp.sm - 1) / 16,
+            })};
+          `}
```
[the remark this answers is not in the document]
mewdriller: From 918px - 991px it has this one-line look that clashes with the design intent. Even below that we end up with the line-breaks not being close to what the design shows. I think it's better off with the `clamp` on the `font-size`; any objections?

image

src/components/util.ts
mewdriller: 🤔 I think it's likely we should be leveraging this in the theme/typography file. We'll need to think through min/max values and the size range those occur in though.

We can also consider using something like `preval` to make these calculations build-time only rather than running the axis-intersection algorithm during the render cycle(s).

tborres: `preval` sounds better to me. There have been a handful of convos about the font sizing for the site: px vs rem, breakpoints vs responsive scaling, etc. No concrete decisions have been made, and there doesn't appear to be a "right" answer. So my typography PR was more of a path forward rather than continuing to stare at the fork in the road 😄

src/components/Header.tsx
```diff
@@ -73,7 +75,7 @@ const A = styled.a(({ theme }) => ({
   ...theme.textVariants.body,
   cursor: 'pointer',
   textDecoration: 'none',
-  ':hover': {
+  '&:hover': {
```
tborres, line 78: 🤔 This seems to work with and without the `&`

src/theme/util.ts
```diff
@@ -0,0 +1,36 @@
...
+
+const relativeBP = {
+  xs: { max: (bp.sm - 1) / 16, min: -Infinity },
+  sm: { max: (bp.md - 1) / 16, min: bp.sm / 16 },
+  md: { max: (bp.lg - 1) / 16, min: bp.md / 16 },
+  lg: { max: (bp.xl - 1) / 16, min: bp.lg / 16 },
+  xl: { max: (bp.xxl - 1) / 16, min: bp.xl / 16 },
+  xxl: { max: (bp.xxl2 - 1) / 16, min: bp.xxl / 16 },
+  xxl2: { max: Infinity, min: bp.xxl2 / 16 },
+};
```
tborres, line 11: Probably worth creating a variable for self-documentation and in case for some reason we were to change the base size.

```suggestion
const BASE_FONT = 16;

const relativeBP = {
  xs: { max: (bp.sm - 1) / BASE_FONT, min: -Infinity },
  sm: { max: (bp.md - 1) / BASE_FONT, min: bp.sm / BASE_FONT },
  md: { max: (bp.lg - 1) / BASE_FONT, min: bp.md / BASE_FONT },
  lg: { max: (bp.xl - 1) / BASE_FONT, min: bp.lg / BASE_FONT },
  xl: { max: (bp.xxl - 1) / BASE_FONT, min: bp.xl / BASE_FONT },
  xxl: { max: (bp.xxl2 - 1) / BASE_FONT, min: bp.xxl / BASE_FONT },
  xxl2: { max: Infinity, min: bp.xxl2 / BASE_FONT },
};
```

tborres (approved):
Ready to merge when you choose - sooner is better than later to reduce the divergency for other branches.

PR #536, opened 2021-04-09
task: footer (B2-9)

k8hansen:
for [B2-9](https://b2io.atlassian.net/browse/B2-9?atlOrigin=eyJpIjoiNDA0NjI1MzJmNjRlNGFhOGEzN2Y5ZWQzNWE2ZmI0ZjMiLCJwIjoiaiJ9)

### **CALLOUTS**

**MAJOR**
We made an executive decision to have the email address break out to the right at a desktop breakpoint, rather than the tablet viewpoint as shown in the mockups. It looked better this way.

**MINOR**

No, the left margin of the footer does not align with the left margin of the header. This is true to the mocks.

**TESTING**
Tested on Chrome, Safari, Firefox, iPhone X, Pixel 5, iPhone 11, Galaxy s21, iPad Pro 2020 & 2017.

Screen Shot 2021-04-09 at 8 44 12 AM


src/components/Footer.tsx
```diff
@@ -0,0 +1,215 @@
...
+          <FooterItem href="/careers">Careers</FooterItem>
+          <FooterItem href="/work">Work</FooterItem>
+          <FooterItem href="/blog">Blog</FooterItem>
+          <FooterItem href="/approach">Approach</FooterItem>
+          <FooterItem href="/contact">Contact</FooterItem>
+          <FooterItem href="/culture">Culture</FooterItem>
+        </ul>
+      </Nav>
+      <ContactBlock>
+        <Link href={'mailto:info@base2.io'}>info@base2.io</Link>
```
k8hansen, line 205: Do we need to do anything with thhe phone number to avoid robocalls? Or to the email address to reduce spam/site crawling?

Do these need an _isExternal_ attribute, similar to the current site?

src/components/Footer.tsx
```diff
@@ -0,0 +1,215 @@
...
+        color: ${(props) => props.theme.colors.coral};
+      }
+    }
+  }
+`;
+
+const ContactBlock = styled.div`
+  color: ${(props) => props.theme.colors.offWhite};
+  font-size: 1.062rem;
+  line-height: 28px;
```
k8hansen, line 134: FYI, no `rem` equivalent given in zepplin.

src/components/Footer.tsx
```diff
@@ -0,0 +1,215 @@
...
+  font-size: 1.875rem;
+  font-weight: normal;
+  line-height: 1.27;
+  letter-spacing: normal;
+  margin: 0.625rem 0 0;
+  width: 10.438rem;
+  word-wrap: auto;
+
+  ${atMinMd} {
+    font-size: 1.875rem;
```
k8hansen, line 56: Woops! Delete.

src/components/Footer.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+    ${atMinLg} {
+      padding-left: 8rem;
+    }
+
+    li {
+      flex-basis: 55%;
+      margin-bottom: 1.25rem;
+
+      &:nth-child(even) {
```
chriswhawkins, line 91: This is a clever way of accomplishing the wrap you're going for, but I think you can really simplify the whole layout by putting your list items [links] in the order you want them to appear and just giving your `ul` a `columns: 2` https://developer.mozilla.org/en-US/docs/Web/CSS/columns

src/components/Footer.tsx
```diff
@@ -0,0 +1,223 @@
...
+  letter-spacing: 0;
+  margin-top: ${({ theme }) => theme.spacing.xxl};
+  padding-bottom: ${({ theme }) => theme.spacing.lg};
+
+  ${atMinMd} {
+    margin-top: 0;
+    padding-bottom: 0;
+  }
+
+  ${atMinLg} {
```
k8hansen, line 131: See pr notes, we decided that this 4th column should happen at a desktop breakpoint, not the tablet as shown in the mockups, for aesthetic reasons.

src/components/Footer.tsx
```diff
@@ -0,0 +1,223 @@
...
+  padding-top: 4rem;
+  position: relative;
+
+  ${atMinMd} {
+    bottom: 0;
+    flex-direction: row;
+    flex-wrap: nowrap;
+    margin-left: 0;
+    padding-bottom: 5rem;
+  }
```
chriswhawkins, line 24: Follow up on the flex comment, I think you can lose all the flex properties on mobile and only apply what you need starting at `atMinMd`...
```
const FooterWrapper = styled.div`
  ...
  ${atMinMd} {
    display: flex;
    justify-content: space-between;
  }
...
```

src/components/Footer.tsx
```diff
@@ -0,0 +1,223 @@
...
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+  margin-left: 1.688rem;
+  max-width: 1600px;
+  padding-top: 4rem;
+  position: relative;
+
+  ${atMinMd} {
+    bottom: 0;
```
chriswhawkins, line 19: Is this necessary?

src/components/Footer.tsx
```diff
@@ -0,0 +1,223 @@
...
+  ${atMinXL} {
+    margin: unset;
+  }
+`;
+
+const Branding = styled.div`
+  display: block;
+
+  ${atMinLg} {
+    width: calc(100% / 3);
```
chriswhawkins, line 35: If this is to divide the width equally among the flex items, `flex: 1` might make more sense

src/theme/colors.ts
```diff
@@ -1,6 +1,8 @@
 export const palette = {
   coral: '#f73a52',
   darkBlue: '#05001e',
+  darkBlueAlt: `#0f0d34`,
+  midBlue: `#18174a`,
```
tborres, line 5: 🔧 Backticks are unnecessary here.
```suggestion
  darkBlueAlt: '#0f0d34',
  midBlue: '#18174a',
```

src/components/index.ts
```diff
@@ -2,6 +2,7 @@ export * from './Container';
 export * from './CTA';
 export * from './Header';
 export * from './Heading';
+export * from './Footer';
```
tborres, line 5: 🔧 Alphabetical order

```suggestion
export * from './Footer';
export * from './Header';
export * from './Heading';
```

src/components/Footer.tsx
```diff
@@ -0,0 +1,223 @@
...
+  }
+`;
+
+const FooterLogo = styled(LogoIcon)`
+  display: block;
+  height: 3.562rem;
+  width: auto;
+
+  ${atMinMd} {
+    height: 3.5rem;
```
chriswhawkins, line 45: My math might be off, but I think .062 of 16px is .... very small. Like less than a pixel.

src/components/Layout.tsx
```diff
@@ -62,44 +64,15 @@ export const Layout: FC<LayoutProps> = ({ children, title = 'Base Two' }) => {
...
-              <a>Blog</a>
-            </Link>
-          </li>
-          <li>
-            <Link href="/contact">
-              <a>Contact</a>
-            </Link>
-          </li>
-        </ul>
+      <Container
```
tborres, line 67: 💬 I think that the `Container` usage should occur within the Footer rather than within the Layout file. That would be consistent with how it is done in the `Header`.

tborres (commented):
First round of tangential comments.

src/components/Footer.tsx
```diff
@@ -0,0 +1,223 @@
...
+  }
+`;
+
+const Slogan = styled.h3`
+  color: ${(props) => props.theme.colors.offWhite};
+  display: inline-block;
```
chriswhawkins, line 51: I don't think this is needed.

src/components/Footer.tsx
```diff
@@ -0,0 +1,223 @@
...
+  font-size: 1.875rem;
+  font-weight: normal;
+  line-height: 1.27;
+  letter-spacing: normal;
```
chriswhawkins, line 55: Is this overwriting a global h3 property?

src/components/Footer.tsx
```diff
@@ -0,0 +1,223 @@
...
+  margin: 0.625rem 0 0;
+  width: 13.062rem;
+  word-wrap: auto;
```
chriswhawkins, line 58: `word-wrap: normal` maybe? I don't think this is a valid property - do we want any special breaking here?

src/components/Footer.tsx
```diff
@@ -0,0 +1,223 @@
...
+    font-size: 1.875rem;
+  }
+`;
+
+const Nav = styled.nav`
+  color: ${(props) => props.theme.colors.offWhite};
+  font-size: ${({ theme }) => theme.spacing.sm};
+
+  ${atMinLg} {
+    width: calc(100% / 3);
```
chriswhawkins, line 70: `flex: 1`

src/components/Footer.tsx
```diff
@@ -0,0 +1,223 @@
...
+  }
+
+  ul {
+    list-style: none;
+    margin: 4rem 0 0 0;
+    padding-left: 0;
+    text-decoration: underline;
+    columns: 2;
```
chriswhawkins, line 78: alpha-ordering is off here - better double check that linting is working and maybe create an issue for getting pre-commit hook for it as well.

src/components/Footer.tsx
```diff
@@ -0,0 +1,223 @@
...
+          <FooterItem href="/work">Work</FooterItem>
+          <FooterItem href="/approach">Approach</FooterItem>
+          <FooterItem href="/culture">Culture</FooterItem>
+          <FooterItem href="/careers">Careers</FooterItem>
+          <FooterItem href="/blog">Blog</FooterItem>
+          <FooterItem href="/contact">Contact</FooterItem>
+        </ul>
+      </Nav>
+      <ContactBlock>
+        <Link
```
tborres, line 203: 🔧 The next `Link` is really for internal navigation. So the usage here and for the telephone should use a standard anchor link (`<a>`)

src/components/Footer.tsx
chriswhawkins: 1 rem = 16px unless we set something different/explicitly on the document

src/components/Footer.tsx
```diff
@@ -0,0 +1,223 @@
...
+  ${atMinMd} {
+    margin-top: 0;
+    padding-bottom: 0;
+  }
+
+  ${atMinLg} {
+    display: flex;
+    flex-direction: column;
+    flex-wrap: wrap;
+    width: calc(100% / 3);
```
chriswhawkins, line 135: `flex: 1`

src/components/Footer.tsx
```diff
@@ -0,0 +1,223 @@
...
+              position: absolute;
+              right: 0;
+            }
+          `}
+        >
+          info@base2.io
+        </Link>
+        <Link href={'tel:6143981158'}>(614) 398.1158</Link>
+        <address>
+          21E 5th Ave. Suite 102
```
tborres, line 216:
```suggestion
          21 E 5th Ave Suite 102
```

src/components/Footer.tsx
```diff
@@ -0,0 +1,223 @@
...
+          <br />
+          Columbus OH 43201
```
tborres, line 218:
```suggestion
          Columbus, OH 43201
```

src/components/Footer.tsx
```diff
@@ -0,0 +1,223 @@
...
+          css={css`
+            ${atMinXL} {
+              position: absolute;
+              right: 0;
+            }
+          `}
+        >
+          info@base2.io
+        </Link>
+        <Link href={'tel:6143981158'}>(614) 398.1158</Link>
```
tborres, line 214:
```suggestion
        <Link href={'tel:6143981158'}>(614) 398-1158</Link>
```

src/components/Footer.tsx
```diff
@@ -0,0 +1,223 @@
...
+export const Footer: FC = () => {
+  return (
+    <FooterWrapper>
+      <Branding>
+        <FooterLogo />
+        <Slogan>Where code meets craft.</Slogan>
+      </Branding>
+      <Nav>
+        <ul>
+          <FooterItem href="/">Home</FooterItem>
```
chriswhawkins, line 193: Maybe `FooterItem` makes more sense as `NavItem` or `FooterNavItem` 🤷‍♂️

src/components/Footer.tsx
```diff
@@ -0,0 +1,223 @@
...
+            ${atMinXL} {
+              position: absolute;
+              right: 0;
+            }
+          `}
+        >
+          info@base2.io
+        </Link>
+        <Link href={'tel:6143981158'}>(614) 398.1158</Link>
+        <address>
```
chriswhawkins, line 215: My understanding of `<address>` is that it is actually for all contact info, so it's kind of like a `<nav>` element in a sense and can provide a content/context wrapper while also giving a free layout/flow wrapper.

src/components/Footer.tsx
```diff
@@ -0,0 +1,223 @@
...
+  }
+`;
+
+interface FooterItemProps {
+  href: string;
+}
+const FooterItem: FC<FooterItemProps> = ({ href, ...props }) => {
+  const { asPath } = useRouter();
+  return (
+    <li className={asPath == href ? 'active' : ''}>
```
tborres, line 176: 💬 It doesn't look like the `.active` class is being targeted anywhere, so it can probably be removed.

```suggestion
  return (
    <li>
```

src/components/Footer.tsx
```diff
@@ -0,0 +1,223 @@
...
+  word-wrap: auto;
+
+  ${atMinMd} {
+    font-size: 1.875rem;
+  }
+`;
+
+const Nav = styled.nav`
+  color: ${(props) => props.theme.colors.offWhite};
+  font-size: ${({ theme }) => theme.spacing.sm};
```
tborres, line 67: ❓ Is this font-size on the `Nav` intentional? Preferably font-sizes should be pulling from or referencing typography rather than spacing.

src/components/Footer.tsx
```diff
@@ -0,0 +1,223 @@
...
+          <FooterItem href="/culture">Culture</FooterItem>
+          <FooterItem href="/careers">Careers</FooterItem>
+          <FooterItem href="/blog">Blog</FooterItem>
+          <FooterItem href="/contact">Contact</FooterItem>
+        </ul>
+      </Nav>
+      <ContactBlock>
+        <Link
+          href={'mailto:info@base2.io'}
+          css={css`
```
chriswhawkins, line 205: I don't mind either approach, but I think there should be a single source of styles - `a:first-of-type` and the `css` helper function are both targeting the same thing, I think.

tborres (commented):
Miscellenous comments, suggestions, and questions. I'm available for a convo now or tomorrow.

Overall great work. It's looking really good.

src/components/Footer.tsx
chriswhawkins: Also, just noticed another alpha-order issue there on the attributes

src/components/Footer.tsx
```diff
@@ -0,0 +1,223 @@
...
+
+  ${atMinLg} {
+    width: calc(100% / 3);
+  }
+
+  ul {
+    list-style: none;
+    margin: 4rem 0 0 0;
+    padding-left: 0;
+    text-decoration: underline;
```
chriswhawkins, line 77: I think it would be better to move any and all `text-decoration` to the actual links. As it is, there are several style blocks and elements to check to make sense of the property.

src/theme/colors.ts
k8hansen: This must be a linter thing, all of the colors in my local have backticks and I'm getting an error when I remove them.

src/components/Layout.tsx
k8hansen: Awesome catch, I didn't notice this! TY

src/components/Footer.tsx
```diff
@@ -0,0 +1,223 @@
...
+import { useRouter } from 'next/router';
+import React, { FC } from 'react';
+import { LogoIcon } from './icons/LogoIcon';
+import { atMinMd, atMinLg, atMinXL } from '~/theme';
+import styled from '@emotion/styled';
+
+const FooterWrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
```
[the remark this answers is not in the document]
k8hansen: I moved` justify-content` to a new breakpoint, it's not needed at mobile.

The other two above styles are needed at mobile. Maybe I'm not understanding exactly what you mean?

src/components/Footer.tsx
k8hansen: I'm not following this one?

src/components/Footer.tsx
k8hansen: Yes, but I forgot to remove some things that I have manually add before we had the variants in! Thanks.

k8hansen: Lol, no. That would be a NJ muscle memory!!

src/components/Footer.tsx
k8hansen: @tborres Tagging you for potential issue creation?

src/components/Footer.tsx
k8hansen: TIL! Refactored this so that the `Contactblock` can be a styled` <address> `tag, let me know if this is not what you had in mind.

src/components/Footer.tsx
```diff
@@ -0,0 +1,207 @@
+import { css, useTheme } from '@emotion/react';
+import styled from '@emotion/styled';
+import Link from 'next/link';
+import React, { FC } from 'react';
+
+import { LogoIcon } from './icons/LogoIcon';
+import { onlyXs, atMinMd, atMinLg } from '~/theme';
+import { Container } from './Container';
+import { Heading } from '~/components';
```
tborres, line 9: This is actually causing a build error because of a circular dependency. Rule of thumb from Drew is to always import peers directly instead of going "up" to the index.
```suggestion
import { Heading } from './Heading';
```

src/components/Footer.tsx
```diff
@@ -0,0 +1,207 @@
...
+            <NavItem href="/approach">Approach</NavItem>
+            <NavItem href="/culture">Culture</NavItem>
+            <NavItem href="/careers">Careers</NavItem>
+            <NavItem href="/blog">Blog</NavItem>
+            <NavItem href="/contact">Contact</NavItem>
+          </ul>
+        </Nav>
+        <ContactBlock>
+          <a href={'mailto:info@base2.io'}>info@base2.io</a>
+          <a href={'tel:6143981158'}>(614) 398-1158</a>
```
tborres, line 197: 🎨 These anchor tags should have the same hover styling as the other anchors. We probably should pull that into a shared styled component (at least within this file for now).

src/components/Footer.tsx
```diff
@@ -0,0 +1,207 @@
...
+          <p>
+            21 E 5th Ave Suite 102
+            <br />
+            Columbus, OH 43201
+          </p>
```
tborres, line 202: 💬 I'm not sure if this _should_ be a `p`, but perhaps we should make it a link to Google maps. Or go retro to MapQuest ;-)

src/components/Footer.tsx
```diff
@@ -0,0 +1,207 @@
...
+      css={css`
+        ${onlyXs} {
+          background-color: ${theme.colors.darkBlueAlt};
+        }
+      `}
+    >
+      <FooterWrapper>
+        <Branding>
+          <FooterLogo />
+          <Slogan as="h3" variant="h3">
```
tborres, line 180: ❓ I wonder, semantically, should this be an `h3` in the markup? It seems like it probably shouldn't be a heading at all 🤷

src/components/Footer.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+      ${atMinMd} {
+        margin-bottom: 1.25rem;
+      }
+    }
+
+    a {
+      color: ${(props) => props.theme.colors.offWhite};
+      font-size: 1.062rem;
+      line-height: 28px;
```
tborres, line 87: Avoid `px`
```suggestion
      line-height: 1.75rem;
```

src/components/Footer.tsx
```diff
@@ -0,0 +1,207 @@
...
+        margin-bottom: 0;
+      }
+
+      ${atMinMd} {
+        margin-bottom: 1.25rem;
+      }
+    }
+
+    a {
+      color: ${(props) => props.theme.colors.offWhite};
```
tborres, line 85: 💬 There's a handful of places where the `props` can be deconstructed into `{ theme }` to be more concise.

tborres (commented):
Looking really good. There's a bit of clean-up that we'll be able to do once we introduce a shared `A` because we'll no longer need some of the color and size overrides in this file.  This is really close to mergeable. There's just a few comments that I'd like to discuss before giving it a thumbs up.

src/components/Footer.tsx
k8hansen: I wondered about this, will do.

src/components/Footer.tsx
k8hansen: Somehow this got reverted back from the first pass, sorry, and fixing.

src/components/Footer.tsx
k8hansen: I see this as being a heading as it is following a branding icon, but I'm fine making it a regular <p> if that makes more sense. I was just trying to be consistent with the mockups.

src/components/Footer.tsx
```diff
@@ -0,0 +1,194 @@
...
+            <NavItem href="/culture">Culture</NavItem>
+            <NavItem href="/careers">Careers</NavItem>
+            <NavItem href="/blog">Blog</NavItem>
+            <NavItem href="/contact">Contact</NavItem>
+          </ul>
+        </Nav>
+        <ContactBlock>
+          <ContactLink href={'mailto:info@base2.io'}>info@base2.io</ContactLink>
+          <ContactLink href={'tel:6143981158'}>(614) 398-1158</ContactLink>
+          <ContactLink href={'https://goo.gl/maps/cesdJy5pkmTqJ7jn7'}>
```
tborres, line 185: I believe that the curly braces are redundant since it's just text links

```suggestion
          <ContactLink href="mailto:info@base2.io">info@base2.io</ContactLink>
          <ContactLink href="tel:6143981158">(614) 398-1158</ContactLink>
          <ContactLink href="https://goo.gl/maps/cesdJy5pkmTqJ7jn7">
```

src/components/Footer.tsx
```diff
@@ -0,0 +1,194 @@
...
+  }
+`;
+
+interface NavItemProps {
+  href: string;
+}
+const NavItem: FC<NavItemProps> = ({ href, ...props }) => {
+  return (
+    <li>
+      <NextLink href={href}>
```
tborres, line 146: 🔧 The `NextLink` is no longer needed in this file, and in fact, this usage is causing some console errors.

src/components/Footer.tsx
```diff
@@ -0,0 +1,194 @@
...
+        </Branding>
+        <Nav>
+          <ul>
+            <NavItem href="/">Home</NavItem>
+            <NavItem href="/work">Work</NavItem>
+            <NavItem href="/approach">Approach</NavItem>
+            <NavItem href="/culture">Culture</NavItem>
+            <NavItem href="/careers">Careers</NavItem>
+            <NavItem href="/blog">Blog</NavItem>
+            <NavItem href="/contact">Contact</NavItem>
```
tborres, line 179: 💬 With the removal of the need for the `NextLink` this area becomes a little less redundant, and it may be more readable to remove the `NavItem` altogether. That's definitely at your discretion though.

```suggestion
            <li>
              <Link href="/">Home</Link>
            </li>
            <li>
              <Link href="/work">Work</Link>
            </li>
            <li>
              <Link href="/approach">Approach</Link>
            </li>
            <li>
              <Link href="/culture">Culture</Link>
            </li>
            <li>
              <Link href="/careers">Careers</Link>
            </li>
            <li>
              <Link href="/blog">Blog</Link>
            </li>
            <li>
              <Link href="/contact">Contact</Link>
            </li>
```

src/components/Footer.tsx
```diff
@@ -0,0 +1,194 @@
...
+    ${atMinMd} {
+      margin-top: 0;
+    }
+  }
+`;
+
+interface NavItemProps {
+  href: string;
+}
+const NavItem: FC<NavItemProps> = ({ href, ...props }) => {
```
tborres, line 143: 💬 If you choose to keep the `NavItem` then the props no longer needs to be deconstructed, and you can pass them wholesale to the `Link`.

tborres (approved):
👯‍♂️

PR #537, opened 2021-04-14
task: Link component

tborres:
closes [B2-68](https://b2io.atlassian.net/browse/B2-68) Anchor component

Introduces a common `Link` component that supports the following variants
- `default`: header & footer links
- `large`: mobile menu links
- `CTA`: call to action links

Changes include
- add `linkVariants` styling to `typography`
- replace CTA usage
- replace next/link usage (aside from the mock examples) with the new `Link` component
- determine internal vs external link based on `href` starting with `/`

Unrelated "bonus" changes
- give header a z-index (based on Material UI's recommended 1100)
- reduce `h3` typography font weight to 300 per conversations with Nelson & Kate

Next steps
- replace downstream branch usages of `a` with the new `Link` component


src/theme/typography.ts
```diff
@@ -1,31 +1,73 @@
...
+    display: 'inline-block',
     fontSize: cssClamp([1.125, 'xs'], [1.375, 'md']),
     fontVariationSettings: '"wght" 900',
     letterSpacing: cssClamp([0.006875, 'xs'], [0.125, 'lg']),
     lineHeight: 1.56,
+    textDecoration: 'none',
+    '::after': {
+      backgroundColor: colors.action,
+      content: '""',
+      display: 'block',
```
k8hansen, line 19: I can't see what this shift between inline-block and block is doing, but I'm going to divert to your expertise and assume that it is needed.

k8hansen (approved):
Might be helpful to have one more set of eyes on the CSS, but overall this is a magical thing of beauty. :shipit:

src/pages/index.tsx
```diff
@@ -3,7 +3,7 @@ import { NextPage } from 'next';
 import { FC } from 'react';
 
 import { atMinLg, cssClamp } from '~/theme';
-import { CTA, Layout, Heading, Text } from '~/components';
+import { Layout, Link, Heading, Text } from '~/components';
```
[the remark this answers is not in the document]
tborres: Does it matter? 
No.

Should it be alphabetical to ensure order and sanity in the world? 
Yes

src/theme/typography.ts
tborres: This `display` value is on the `after`, which is for the red line. The `display: 'inline-block'` is on the root `a`, which was applied to ensure that the item doesn't take the full width of the screen thereby stretching out the red line (which is what it would do if it were `block` or `inline`).

k8hansen: thanks!

PR #540, opened 2021-04-17
feat(home) home hero animations

mewdriller:
Update the Home Hero to included the animated SVG lines.

[B2-11](https://b2io.atlassian.net/browse/B2-11)


tborres (commented):
Just a couple of comments so far. It looks good. Lots of magic numbers, which is to be expected, but calls out more of a reason to extract to a separate file.

Aesthetically I'm not sure where we want to go with this. I don't have the best eye for this type of thing.

src/pages/index.tsx
```diff
@@ -1,11 +1,93 @@
...
+
+interface LineProps {
+  index: number;
+  threshold: number;
+}
+
+const Line: FC<LineProps> = ({ index, threshold }) => {
+  const pathRef = useRef<SVGPathElement>(null);
+  const { scrollY } = useViewportScroll();
+  // [0, threshold] => [-1, 1]
```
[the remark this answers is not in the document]
mewdriller: It's supposed to provide some context that we're mapping numbers in the `[0, threshold]` space to the `[-1, 1]` space. I can remove/expand it if you feel like it's unhelpful.

src/components/HomePage/HomeHero.tsx
```diff
@@ -0,0 +1,168 @@
...
+      css={css`
+        display: grid;
+        gap: ${cssClamp([2, 'xs'], [1.5, 'lg'])};
+        padding: ${cssClamp([3, 'xs'], [10.5, 'lg'])}
+          ${cssClamp([1.5, 'xs'], [5.25, 'lg'])};
+        position: relative;
+        text-align: center;
+
+        & * {
+          text-shadow: 0.25rem 0.25rem 0.375rem #05001e;
```
chriswhawkins, line 103: Is this a "one off" color or something that we could get from the theme?

mewdriller: _Definitely_ not a one-off color. Good call!

src/components/HomePage/HomeHero.tsx
```diff
@@ -0,0 +1,168 @@
...
+      stroke="transparent"
+      strokeWidth="0.0125em"
+      style={{ opacity }}
+      width="100%"
+    />
+  );
+};
+
+export const HomeHero: FC = () => {
+  const [ref, { bottom }] = useMeasure<HTMLDivElement>();
```
chriswhawkins, line 90: Would `bottom` be equal to `height` for the ref'd element? Mostly just curious, b/c I really don't know, but also if it is, `height` might be a little more intuitive.

mewdriller: Hmmm... I _think_ it's the equivalent of `top + height`.

src/components/HomePage/HomeHero.tsx
mewdriller: I got rid of the magic color here, and also in the linear-gradient. That led to figure out how to smoothly interpolate n-stop linear gradients (which was pretty fun).

It should remove all the magic colors and make it change if we ever tweak the theme values.

chriswhawkins:
Everything looks really awesome.

I might try to riff on this branch a bit tonight and potentially circle back on this comment, but wanted to leave a couple of first impressions/ideas in case anyone else gets a chance to takes a look and has similar or countering opinions:

1. I wonder if the lines would look better going full bleed or with a softened/feathered edge.
2. The "lighting" animation is awesome, but so subtle that I wonder if it's easy to miss. Speeding it up and/or mirroring it on repeat might bring it out a bit more, but I do think subtle is sort of the goal.
3. I love the transform with the scroll and I'm wondering if we can ratchet it up a notch. Maybe adding even more vertical whitespace below so that the transform can be more exaggerated. I would imagine adding a gradient overlay on the bottom, too, like we get on top from the menu.

src/components/HomePage/HomeHero.tsx
```diff
@@ -0,0 +1,169 @@
...
+      stroke="transparent"
+      strokeWidth="0.0125em"
+      style={{ opacity }}
+      width="100%"
+    />
+  );
+};
+
+const STOP_COLORS = interpolateColors(
+  [theme.colors.midBlue, theme.colors.coral, theme.colors.midBlue],
```
chriswhawkins, line 90: love this 👏

src/components/HomePage/HomeHero.tsx
```diff
@@ -0,0 +1,169 @@
...
+      `C1000,${y(4)} 1000,${y(4)} ${x(875)},${y(4)}`,
+      `C${x(750)},${y(4)} ${x(625)},${y(peakY + 4)} ${x(500)},${y(peakY + 4)}`,
+      `C${x(375)},${y(peakY + 4)} ${x(375)},${y(4)} ${x(250)},${y(4)}`,
+      `L0,${y(4)}`,
+      `Z`,
+    ].join(' ');
+  });
+
+  // Given the scrollYRange value and the index, compute the opacity for this curve.
+  const opacity = useTransform(scrollYRange, (value) => {
```
FrankyMartz, line 67: **Opacity**

On initial visit noticed a grid flicker, and how it snaps on my normal scroll. At first I thought it was a pattern flickering (from grid) but then also noticed the opacity fade. I have some concerns here because it's really tricky getting a smooth fade, and most people don't scroll slowly.

Perhaps we should animate it separate from scrolling.

If we are really set on an animated fade, perhaps just one direction at the moment it appears to fade-out-in.

**Motion**

Because the scrolling area is so small it sort of snaps when I scroll.
Only solutions I can think of here is triggering animation with small number of breakpoints. The concern here is maintaining smooth transitions which we might not achieve by directly binding to scroll.

FrankyMartz (commented):
Noted some concerns about flickering and snapping on the `HomeHero.tsx`. Please let me know if you need more insight.

src/components/HomePage/HomeHero.tsx
mewdriller: Yeah, I noticed that too, it's definitely a little tight basing the animation off  solely the scroll position. We can use a [cycle](https://www.framer.com/api/motion/utilities/#usecycle) or some other user-input based approach (relative mouse position, etc).

I'm open to any all direction at this point, but I think the biggest thing is we need to decide and go. This is _definitely_ one of those cards that can float around forever in tweaks and edits.

src/components/HomePage/HomeHero.tsx
```diff
@@ -0,0 +1,246 @@
...
+        position: relative;
+        text-align: center;
+
+        & * {
+          text-shadow: 0.25rem 0.25rem 0.375rem ${theme.colors.darkBlue};
+          z-index: 1;
+        }
+
+        & svg {
+          z-index: 0;
```
tborres, line 198: 💬 I think that this would look better stretched edge to edge. When I slapped the following into Chrome dev tools it seemed to work, but it may not be the "right way" to do it.

```suggestion
          left: calc(50% - 50vw);
          width: 100vw;          
          z-index: 0;
```

tborres (approved):
So much math!

The animation is looking really good! I think that it would look better with a full bleed edge to edge or a fading on the edges, but that can be addressed in a future branch unless it's a quick change.

With the folder structure changes and the introduction of `polished` and the `interpolateColors` util it would be good to get this merged sooner rather than later - i.e. this morning.

src/components/HomePage/BuiltForResults.tsx
```diff
@@ -132,5 +132,3 @@ export const BuiltForResults: FC = () => {
     </section>
   );
 };
-
-export default BuiltForResults;
```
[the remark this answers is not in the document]
mewdriller: My understanding of the suggested best practices in the TypeScript world is that named exports are preferred to default exports as the play nicer with import type-sense + refactoring as well as the `export * from './module';` syntax.

@FrankyMartz  + @sludemann: Any thoughts on ☝️ from your more extensive TS explorations?

src/components/HomePage/HomeHero.tsx
```diff
@@ -0,0 +1,246 @@
...
+    const peakY = LINE_APEX_Y * mY * yMod;
+    const yValue = transform(index, LINE_RANGE, [0, LINE_WIDTH - LINE_HEIGHT]);
+    const y = (v: number): number => v + yValue;
+
+    return [
+      `M0,${y(0)}`,
+      `L${x(x1)},${y(0)}`,
+      // prettier-ignore
+      `C${x(x1 + SEGMENT_WIDTH)},${y(0)} ${x(x2 - SEGMENT_WIDTH)},${y(peakY)} ${x(x2)},${y(peakY)}`,
+      // prettier-ignore
```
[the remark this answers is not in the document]
mewdriller: Yep. It _definitely_ does. I spent *way* too long trying to figure that out but they don't seem to have the comment pragmas you would expect coming from `eslint` and similar libraries.

If you (or anyone else) know the magic spell, please let me know. 🙃

src/components/HomePage/HomeHero.tsx
mewdriller: 🏆 Excellent suggestion. I went ahead and incorporated that and pushed.

src/components/HomePage/BuiltForResults.tsx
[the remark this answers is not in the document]
sludemann: I lean away from default exports mostly for consistency. Library code in general doesn't wind up going for default exports for obvious reasons and I just find it messy to mix the styles of imports / exports.  Honestly I think having the distinction was a fundamentally bad design choice and everything should have just been named imports / exports in general.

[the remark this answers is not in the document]
FrankyMartz: Hmm, I haven't heard of `default` assignment negatively affecting type-sense. There is some conscious overhead with re-exporting, however you also gain the ability of simplified import renaming. Pros and cons to both I suppose.

Personally, I do find myself using the `default` assignment in my own work; since, it forces me to think of modules singular in purpose and with ideally one export. However, the single export module is rarely the case in real world production, but I do appreciate the enforced reminder.

This is all entirely a reasoning of preference; so, I have no concerns with strictly using named exports.

src/components/HomePage/HomeHero.tsx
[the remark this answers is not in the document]
tborres: I had spent 10-15 minutes trying to figure it out and then felt that I was wasting too much time on it. It works as it is 👍

PR #543, opened 2021-04-20
chore: linting

tborres:
Closes [B2-69](https://b2io.atlassian.net/browse/B2-69) Alpha Linting

- add `stylelint` and pre-commit husky hook
- add alphabetical rules to `eslint`
- update files that failed the new linting rules

### Dev Setup / Usage Notes
It is recommended that you install the [`stylelint` VS Code extension](https://marketplace.visualstudio.com/items?itemName=stylelint.vscode-stylelint) so that you get warnings within your IDE. If you find the extension throwing warnings for unrelated files (such as markdown files) you can add a global `.stylelintrc` file to your personal root directory (i.e. the same location as your `.gitconfig`) that includes the following:

```
{
   "ignoreFiles": ["**/*.md"]
}
```


tborres:
Stylelint config coming soon...

src/theme/typography.ts
```diff
@@ -26,115 +18,115 @@ export const linkVariants = {
...
+    fontSize: cssClamp([1.125, 'mobile'], [1.375, 'tablet']),
+    fontVariationSettings: '"wght" 900',
+    letterSpacing: cssClamp([0.006875, 'mobile'], [0.125, 'lg']),
+    lineHeight: 1.56,
+    textDecoration: 'none',
   },
   default: {
+    '&:hover': {
+      color: colors.coral,
+    },
```
chriswhawkins, line 37: The diff is a little hard to follow and I'm multi-tasking, but I think I would expect to see specificity (use cases and overwrites) to be in a descending order, i.e.: initial properties, then pseudo, then targets, then breakpoints

src/theme/typography.ts
```diff
@@ -26,115 +18,115 @@ export const linkVariants = {
...
-
-    [atMinTablet]: {
-      letterSpacing: 'normal',
-      lineHeight: 1.2,
-    },
   },
   h1: {
+    [atMinTablet]: {
+      lineHeight: 1.04,
+    },
```
chriswhawkins, line 83: same comment as above, this might not be causing issues in how the css is actually compiled, but I think it's more intuitive to list base properties before more specific or overwriting use cases like breakpoints.

tborres: 100% agree that the original order is more intuitive, but this is a consequence of the alpha-sorting rules. I'm not aware of a way around it other than not applying the linting at all (or using a lot of ignores).

.stylelintrc
```diff
@@ -0,0 +1,16 @@
...
+  "extends": [
+    "stylelint-config-recommended",
+    "stylelint-config-styled-components"
+  ],
+  plugins: ['stylelint-order'],
+  "processors": [
+    "stylelint-processor-styled-components"
+  ],
+  rules: {
+    'no-descending-specificity': null,
```
chriswhawkins, line 11: If you turn this rule on, does it conflict with the alpha ordering?

src/theme/typography.ts
tborres: This is due to the `eslint` rules since this is using the object-notation of styled components. This "unnatural" ordering will not be applied for the css-syntax versions.

tborres:
After a bit of conversation with Chris and Nelson I'm going to go ahead and merge this linting PR. We can always make changes later.

PR #544, opened 2021-04-20
task: featured case study pane

nelsW70:
Closes [B2-60](https://b2io.atlassian.net/browse/B2-60)

- Created component within pages/home folder to contain the text and image related to the Featured case study pane
- Laid out static text for all screen sizes
- Edited existing png image assets to be used as temp images

**Next Steps**
- [ ] acquire/replace the image assets
- [ ] add box shadow, add border-radius to image container divs
- [ ] add background image once acquired from designers

Note: The "As a leader..." font styles differ from the design because it was discussed and the decision was to not change it once it reaches tablet.


src/components/pages/home/FeaturedCaseStudy.tsx
```diff
@@ -0,0 +1,152 @@
...
+  ${atMinXL} {
+    margin-top: 32%;
+    order: 2;
+    width: calc(50% - ${GUTTER_SHIFT});
+  }
+`;
+
+const ImagesContainer = styled.div`
+  columns: 2;
+  column-gap: ${theme.spacing.xxs};
```
chriswhawkins, line 35: @tborres calling this out because we just had this gap support conversation last week - Oddly enough, in this context (multi-column layout, vs flex) it _is_ supported in Safari except when percent or calc() values.

src/components/pages/home/FeaturedCaseStudy.tsx
```diff
@@ -0,0 +1,152 @@
...
+          `}
+          href="/work"
+          variant="CTA"
+        >
+          See our work
+        </Link>
+      </TextContainer>
+      <ImagesContainer>
+        <ResponsiveImage
+          alt="case study image"
```
tborres, line 119: 🔧 The alt tag should be a bit more descriptive of the picture and does not need to include the term "image" because it's already known that it is an image. For this one the alt tag could be `paint a photo` 🤷‍♂️

src/components/pages/home/FeaturedCaseStudy.tsx
```diff
@@ -0,0 +1,152 @@
+import { css } from '@emotion/react';
+import styled from '@emotion/styled';
+import NextImage from 'next/image';
+import { FC } from 'react';
+
+import { Heading, Link, Text } from '~/components';
+import theme, { atMinSm, atMinTablet, atMinXL, cssClamp } from '~/theme';
+
+type ImageProps = {
```
tborres, line 9: 💬 I might put this `ImageProps` definition just before the `ResponsiveImage` because they're related.

src/components/pages/home/FeaturedCaseStudy.tsx
```diff
@@ -0,0 +1,152 @@
...
+  alt: string;
+  src: string;
+};
+
+const GUTTER_SHIFT = '1.375rem';
+
+const FeaturedCaseStudyWrapper = styled.section`
```
tborres, line 16: 💬 I go back and forth on this type of thing so much, so large grain of salt here, but since this is such a simple styling it may be more readable to nix the named styled and just have a `section` with this `css`.

src/components/pages/home/FeaturedCaseStudy.tsx
```diff
@@ -0,0 +1,152 @@
...
+  columns: 2;
+  column-gap: ${theme.spacing.xxs};
+  ${atMinTablet} {
+    column-gap: 2.875rem;
+  }
+  ${atMinXL} {
+    width: calc(50% + ${GUTTER_SHIFT});
+  }
+  > div {
+    break-inside: avoid;
```
tborres, line 43: ❓ Why is this needed?

src/components/pages/home/FeaturedCaseStudy.tsx
```diff
@@ -0,0 +1,152 @@
...
+  flex-direction: column;
+  ${atMinXL} {
+    flex-direction: row;
+  }
+`;
+
+const TextContainer = styled.div`
+  margin-left: ${cssClamp([1.74375, 'sm'], [5.625, 'tablet'], [7.6125, 'xl'])};
+  ${atMinXL} {
+    margin-top: 32%;
```
tborres, line 27: 💬 Since most of the rest of our positioning and sizing are rem-based I wonder if this should be `rem` rather than `%`? Honestly, I'm not sure though.

src/components/pages/home/FeaturedCaseStudy.tsx
```diff
@@ -0,0 +1,152 @@
...
+          Visualizer
+        </Heading>
+        <Text
+          variant="h3"
+          css={css`
+            max-width: 19rem;
+            ${atMinSm} {
+              max-width: 30rem;
+            }
+          `}
```
tborres, line 93: 💬 While this is true to the design, I wonder if we need the max-width on this text at all.

tborres (commented):
Looking really good. I have handful a questions/comments, but we should be able to get this merged today.

src/components/pages/home/FeaturedCaseStudy.tsx
chriswhawkins, line 43: Curious what this is doing. Does this `break-inside` prevent some part of the Next Image component from wrapping before/without the actual image tag?

src/components/pages/home/FeaturedCaseStudy.tsx
```diff
@@ -0,0 +1,152 @@
...
+        >
+          As a leader in the paint industry, Sherwin-Williams wanted to create a
```
chriswhawkins, line 95: Perhaps this is identified in another card, but curious if there's a plan to have content like this lifted out of what might otherwise be a reusable component. (@tborres)

src/components/pages/home/FeaturedCaseStudy.tsx
```diff
@@ -0,0 +1,152 @@
...
+          tool worthy of their products - something to get customers excited
+          about paint.
+        </Text>
+        <Link
+          css={css`
+            margin-top: ${theme.spacing.lg};
+            margin-bottom: ${theme.spacing.lg};
```
chriswhawkins, line 102: alpha order

src/components/pages/home/FeaturedCaseStudy.tsx
chriswhawkins, line 119: maybe these are placeholders, but just in case, these alt values will need some work. I don't have the comp handy, but can help work out what we might want to omit as more presentational vs what we can label a little more clearly.

chriswhawkins (commented):
👍 🏅 Nice work! I only did a quick (distracted) run through of the diff, but code is looking good! I had a few comments and questions we can follow up on whenever there's time (or feel free to disregard/ignore as needed)

src/components/pages/home/FeaturedCaseStudy.tsx
```diff
@@ -0,0 +1,152 @@
...
+        width={400}
+        height={706}
+      />
+    </div>
+  );
+};
+
+export const FeaturedCaseStudy: FC = () => {
+  return (
+    <FeaturedCaseStudyWrapper>
```
asingleton6, line 63: @tborres @nelsW70
Will white space be added to the top and bottom of each section?

src/components/pages/home/FeaturedCaseStudy.tsx
nelsW70: @tborres and @chriswhawkins, so is it safe to use column-gap here?

src/components/pages/home/FeaturedCaseStudy.tsx
```diff
@@ -0,0 +1,152 @@
...
+          href="/work"
+          variant="CTA"
+        >
+          See our work
+        </Link>
+      </TextContainer>
+      <ImagesContainer>
+        <ResponsiveImage
+          alt="case study image"
+          src="/home/img-sw-home-1@3x_TEMP.png"
```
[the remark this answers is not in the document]
nelsW70: @tborres, how about "featured-case-study-home-1", ""featured-case-study-home-2" etc.?

src/components/pages/home/FeaturedCaseStudy.tsx
nelsW70: @tborres, good call. I should have known that, dumb oversight on my part.

src/components/pages/home/FeaturedCaseStudy.tsx
nelsW70: @tborres, it is in fact not needed. I had used it earlier to prevent the image from breaking but changed the approach and it's no longer necessary.

src/components/pages/home/FeaturedCaseStudy.tsx
```diff
@@ -0,0 +1,152 @@
...
+  > div {
+    break-inside: avoid;
+  }
+`;
+
+const ResponsiveImage: FC<ImageProps> = ({ alt, src, ...props }) => {
+  return (
+    <div {...props}>
+      <NextImage
+        layout="responsive"
```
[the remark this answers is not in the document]
nelsW70: @chriswhawkins, would the alpha order rule apply to the css helper as well? 

Like this?

as="h3"
css={css`
            margin-bottom: ${theme.spacing.sm};
          `}
variant="h2"

src/components/pages/home/FeaturedCaseStudy.tsx
nelsW70: @chriswhawkins, it was wrapping so that was to prevent that behavior but I updated the approach to the Next Image component and so it's no longer necessary. Thanks for catching that!

src/components/pages/home/FeaturedCaseStudy.tsx
nelsW70: @chriswhawkins, I should have caught that. It was an oversight on my part and I added more descriptive and unique alt text.

src/components/pages/home/FeaturedCaseStudy.tsx
nelsW70: @asingleton6, from what I understand that layout styling will be implemented in a different task

src/components/pages/home/FeaturedCaseStudy.tsx
```diff
@@ -0,0 +1,152 @@
...
+type ImageProps = {
+  alt: string;
+  src: string;
+};
+
+const GUTTER_SHIFT = '1.375rem';
+
+const FeaturedCaseStudyWrapper = styled.section`
+  display: flex;
+  flex-direction: column;
```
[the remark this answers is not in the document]
nelsW70: @chriswhawkins

Yup. Thanks!

src/components/pages/home/FeaturedCaseStudy.tsx
nelsW70: @tborres Sure, makes sense to go with consistency

src/components/pages/home/FeaturedCaseStudy.tsx
nelsW70: @tborres Might want to leave it for now. Otherwise, letting it break naturally causes Sherwin-Williams to wrap which seems like a bad thing. Totally your call of course

src/components/pages/home/FeaturedCaseStudy.tsx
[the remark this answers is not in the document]
chriswhawkins: Yeah, exactly. Not sure if that's everyone else's preference, though. Not sure if it's included in the linting work (@tborres), but I know there's an eslint flag for this with additional options.

tborres (approved):
🍨

src/components/pages/home/FeaturedCaseStudy.tsx
asingleton6: Got it! Thanks for clearing that up.

PR #545, opened 2021-04-21
task: built for results

nelsW70:
Closes [B2-59](https://b2io.atlassian.net/browse/B2-59)

- Created a component within the Home page folder to contain the text and image related to the Built for Results pane
- Mirrored Tony's approach to Designed for Humans pane
- Adjusted styling as needed for layout variation

**Next Steps**
- [ ] ensure uniformity between panes on home page
- [ ] might want to consider a variant for the Link component as the pseudo after margin-top is different on the Home page designs than it is on the mobile menu: just a small tweak but should probably be addressed

<details>
<summary>mobile</summary>

Screen Shot 2021-04-21 at 10 46 12 AM
</details>

<details>
<summary>tablet</summary>

Screen Shot 2021-04-21 at 10 46 30 AM
</details>

<details>
<summary>desktop</summary>

Screen Shot 2021-04-21 at 10 46 48 AM
</details>


src/components/pages/home/index.tsx
```diff
@@ -1,14 +1,16 @@
 import { FC } from 'react';
 
 import { Layout } from '~/components';
-import DesignedForHumans from './DesignedForHumans';
 import HomeHero from './HomeHero';
+import DesignedForHumans from './DesignedForHumans';
```
tborres, line 5: 🔧 💬 By our general convention these imports should be in alphabetical order based on file location (e.g. Built, Designed, and then HomeHero) even though their usage within the page is different.

src/components/pages/home/BuiltForResults.tsx
```diff
@@ -0,0 +1,143 @@
...
+import { Heading, Link, Text } from '~/components';
+import theme, {
+  atMinSm,
+  atMinTablet,
+  atMinLg,
+  atMinXL,
+  cssClamp,
+} from '~/theme';
+
+type ImageProps = {
```
tborres, line 15: 💬 We've defined this in a few places. We should consider lifting this up to a `/src/types.ts` file.

src/components/pages/home/BuiltForResults.tsx
```diff
@@ -0,0 +1,143 @@
...
+import NextImage from 'next/image';
+import { FC } from 'react';
+
+import { Heading, Link, Text } from '~/components';
+import theme, {
+  atMinSm,
+  atMinTablet,
+  atMinLg,
+  atMinXL,
+  cssClamp,
```
tborres, line 12: 💬 The linter will eventually catch this, but these should be in alpha order

tborres (approved):
This looks great. A couple of things to potentially adjust before merging, but otherwise this is good to go.

To what extent did you consider creating a shared component between this and the Designed For Humans pane? Did it seem like the position was different enough that it wouldn't be worth it?

- [ ] The `img-home-`... files are not being used and therefore should be removed.
- [ ] We've now seen the `ImageProps` type defined a few times. We should refactor it to a shared type. This could be a separate task
- [ ] There is probably a few linting issues that will be caught once the linter is introduced.

src/components/pages/home/BuiltForResults.tsx
nelsW70: @tborres Sure, just added it and implemented it in the DesignedForHumans and BuiltForResults component

nelsW70:
@tborres All the changes you suggested are in

Should we discuss making the shared component for this and Designed For Humans pane before merging it in or after? 

I did think about the image part of it being reusable and it should be possible

PR #546, opened 2021-04-26
B2 71 add gradient to the Feature Case Study widget

asingleton6:
Closes [B2-71](https://b2io.atlassian.net/browse/B2-71)
- Add gradient to the Feature Case Study widget.

image


src/components/pages/home/FeaturedCaseStudy.tsx
```diff
@@ -58,7 +81,12 @@ export const FeaturedCaseStudy: FC = () => {
         }
       `}
     >
-      <TextContainer>
+      <TextContainer
```
nelsW70, line 84: @asingleton6, I'd probably move these two rules into the styled component itself to keep things clean but not sure what the seniors think

Also, not sure how others approach this but I'd rather see z-index values be as low as possible and only as much as is necessary to increase hierarchy over other elements in the DOM. Generally increasing in single increments.

So like, element A has z-index: 1, element B has z-index: 2, etc.  Otherwise, you get into the z-index: 1,000,000 real quick

src/components/pages/home/FeaturedCaseStudy.tsx
```diff
@@ -16,16 +16,39 @@ const TextContainer = styled.div`
...
+    rgba(45, 45, 120, 1) 80%,
+    rgba(22, 21, 69, 1) 88%,
+    rgba(4, 0, 27, 1) 91%
+  );
+  bottom: 0;
+  height: 123%;
+  left: -1.5rem;
+  position: absolute;
+  top: -4rem;
+  width: calc(100% + (2 * 1.5rem));
```
tborres, line 34: 🔧 There's a handful of places in this `Gradient` that have spacing values that match the `spacing.ts` file. Consider leveraging the theming in those cases.

e.g. (note that you'll need to `import { spacing } from '~/theme/spacing';` for this usage.

```suggestion
  width: calc(100% + (2 * ${spacing.sm}));
```

src/components/pages/home/FeaturedCaseStudy.tsx
tborres: I agree with Nelson. z-index value picking is always fun. Generally I think that it should be as low as possible unless it has a functional reason to be a larger number (such as for a header or modal windows).

tborres (commented):
The gradient looks pretty good. The Zeplin version seems to have some sort of noise to it, but I'm not sure how that would be applied. I'd like to get @chriswhawkins eyes on it from a visual and CSS standpoint. I'm not sure if any of those rgb colors match up to our color palette. If so then we'll when to drive those off the theme.

In addition to the comments that I added to the source I think that we'll need to introduce a sizing mechanism to ensure that the contents of gradient don't bleed outside of the pane's container. We could add more spacing on the home page, but in general I've been thinking of it as though each pane should fully contain itself (unless it's visuals need to overlap other pane content).

Aside from the source code changes, take a look at the [Development Workflow](https://github.com/b2io/base2.io/tree/next#readme) in the next branch's README when you get a chance.. With Jira being newish for all of us I detailed a suggested card management and PR process there.

src/components/pages/home/FeaturedCaseStudy.tsx
asingleton6: Changes have been merged!

src/components/pages/home/FeaturedCaseStudy.tsx
```diff
@@ -10,22 +10,47 @@ const GUTTER_SHIFT = '1.375rem';
...
+  z-index: 1;
   ${atMinXL} {
     margin-top: 22rem;
     order: 2;
     width: calc(50% - ${GUTTER_SHIFT});
   }
 `;
+// The gradient will overlap the "Built for Results" component, on desktop, until padding is added between each sections
+const Gradient = styled.div`
+  background: linear-gradient(
```
mewdriller, line 23: 🤔 This might be a candidate for either [`polished`'s `linearGradient` helper](https://polished.js.org/docs/#lineargradient) or the `interpolateColors` helper on my branch (although that only currently works on colors w/o an alpha channel).

That would allow us to use `theme` colors for the start/stop and compute the between steps.

src/components/pages/home/FeaturedCaseStudy.tsx
```diff
@@ -10,22 +10,47 @@ const GUTTER_SHIFT = '1.375rem';
...
   margin-left: ${cssClamp([1.74375, 'sm'], [5.625, 'tablet'], [7.6125, 'xl'])};
+  position: relative;
+  z-index: 1;
   ${atMinXL} {
     margin-top: 22rem;
     order: 2;
     width: calc(50% - ${GUTTER_SHIFT});
   }
 `;
+// The gradient will overlap the "Built for Results" component, on desktop, until padding is added between each sections
```
mewdriller, line 21: ❓ Do we have a follow-up card to remove this once the trigger condition ("padding is added...") is met?

src/components/pages/home/FeaturedCaseStudy.tsx
```diff
@@ -10,22 +10,47 @@ const GUTTER_SHIFT = '1.375rem';
 
 const TextContainer = styled.div`
   margin-left: ${cssClamp([1.74375, 'sm'], [5.625, 'tablet'], [7.6125, 'xl'])};
+  position: relative;
+  z-index: 1;
```
mewdriller, line 14: ⚠️ There's some general _issues_ with the `z-index` organization in the site at this point. I think we need to align on some "depths" for things to sit at by default so we can add these background elements without colliding with content above/below.

image

mewdriller: 🤔 Also, (totally subjectively) the horizontal edges of that gradient look *sharp* to me. Seems like _maybe_ it should have a bit of blending on those edges as well? That could be just a biproduct of the overlap though. 🤷

asingleton6: @mewdriller @tborres 
I updated the gradient to be contained inside the parent container on desktop! Do the edges still look too sharp??

Screen Shot 2021-04-27 at 11 01 04 AM

src/components/pages/home/FeaturedCaseStudy.tsx
asingleton6: I'll add a card in Jira to add this. Thanks Drew!

tborres (changes requested):
It looks like the gradient is escaping the container on the bottom when between a mobile and tablet breakpoint (e.g. 700px). This may be due to the `height: 123%`, but I'm not sure. May need to add a calculated min-height to ensure that it doesn't collide with other panes.

Once the sizing issue is addressed it should be good to merge and then begin work on the home page layout task while improving the smoothness of this gradient in a separate task.

image

asingleton6:
@mpcondo 
Thank you!! ☺️  

I noticed in the mockup that the gradient actually flows past the picture, throwing off the balance. 
https://zpl.io/V1RyNQB

Screen Shot 2021-04-28 at 1 47 18 PM

nelsW70:
Looks good to me! I think Tony's comment about smoothing the gradient blend out in another task makes sense. That's really the only thing I notice that hasn't already been addressed. 

Nice work Ashley!

tborres (approved):
Nice job 👍

PR #547, opened 2021-04-26
B2 61 fueled by curiosity

nelsW70:
Closes [B2-61](https://b2io.atlassian.net/browse/B2-61)

- Created a component within the Home page folder to contain the text and image related to the Fueled By Curiosity pane
- Added "xxxs" variable in spacing.ts and implemented it in Link component to unify CTA styling 

NOTES: 

- If "xxxs" is garbage-naming for the spacing var just yell and I'll make it something else
- Tried to make the the large circle gradient resizing as clean/simple as possible knowing that it may be replaced entirely when the animation is implemented
- The rectangular transparent gradient present in the design was not added after a discussion with team members

**Next Steps**
- [ ] add large circle gradient animation

<details>
<summary>mobile</summary>

Screen Shot 2021-04-26 at 10 20 09 AM
</details>

<details>
<summary>tablet</summary>

Screen Shot 2021-04-26 at 10 20 29 AM

</details>

<details>
<summary>desktop</summary>

Screen Shot 2021-04-26 at 10 21 04 AM

</details>


src/theme/spacing.ts
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable sort-keys */
 export const spacing = {
+  xxxs: '.25rem', // 4px
```
tborres, line 3: Sigh. I do hate the naming, but I don't have a naming convention that I don't hate. We could name it like the xxl versions, but I didn't like that approach either.

E.g. 
```suggestion
  xxs2: '.25rem', // 4px
```

tborres (approved):
Looks great. Just may want to come up with a better spacing naming convention.

src/theme/spacing.ts
nelsW70: Yeah, it did not feel good at all to type that out. Not the least bit proud of it. 

Following the xxl approach is less lame but more confusing in my opinion because the integer could be interpreted as smaller or larger

I mean, we could go with "tiny" or "smallest" but those break the pattern too.

src/components/pages/home/FueledByCuriosity.tsx
```diff
@@ -0,0 +1,105 @@
+import { css } from '@emotion/react';
+import styled from '@emotion/styled';
+import NextImage from 'next/image';
+import { FC } from 'react';
+
+import { Heading, Link, Text } from '~/components';
+import theme, { atMinTablet, atMinXL, cssClamp } from '~/theme';
+import { ImageProps } from '~/types';
+
+const imageDimensionsCalc = cssClamp([32, 'mobile'], [69.5, 'tablet']);
```
FrankyMartz, line 10: Why are you avoiding Media queries?
I have some serious concerns about using clamp in this manner. Instead of dimensions calculated in chunks they are now evaluated at each level for the smallest possible components. Where as media queries allow the browser to disregard blocks of styling where conditions aren't met.

Clamps are useful however feel like we should be mindful of when they are used. Since html elements can achieve same fluid behavior with just a relative unit — i.e. percentages or viewport units

src/components/pages/home/FueledByCuriosity.tsx
```diff
@@ -0,0 +1,105 @@
...
+        right: calc(77.5% - ${imageDimensionsCalc});
+        top: ${imageTopCalc};
+        width: ${imageDimensionsCalc};
+        ${atMinXL} {
+          right: -22rem;
+        }
+      `}
+      {...props}
+    >
+      <NextImage layout="fill" alt={alt} src={src} />
```
FrankyMartz, line 28: We should really be using `sizes` here. The Zeplin application allows use to export different sizes for each medium.

src/components/pages/home/FueledByCuriosity.tsx
tborres: @FrankyMartz We've been using this clamp approach throughout the home page tasks. This is definitely something that we can talk about, but it would be a larger conversation outside of this task.

src/components/pages/home/FueledByCuriosity.tsx
tborres: As I understand it the `NextImage` will handle the responsive sizing on-demand, which is why we're not using the Zeplin sizes.

src/components/pages/home/FueledByCuriosity.tsx
```diff
@@ -0,0 +1,105 @@
...
+  ${atMinTablet} {
+    margin-left: ${cssClamp([0, 'tablet'], [7.6125, 'xl'])};
+  }
+`;
+
+export const FueledByCuriosity: FC = () => {
+  return (
+    <section
+      css={css`
+        min-height: calc(${imageTopCalc} + ${imageDimensionsCalc});
```
FrankyMartz, line 71: It's usually best to avoid limiting dimensions or setting a minimum dimension. In this instance, I imagine the content is sufficient to occupy enough vertical content to correctly display the background gradient circle.

src/components/pages/home/FueledByCuriosity.tsx
```diff
@@ -0,0 +1,105 @@
...
+    ${atMinTablet} {
+      top: -1.25rem;
+    }
+  }
+`;
+
+const Content = styled.div`
+  margin-left: 1.74375rem;
+  margin-top: ${cssClamp([6.75, 'mobile'], [19, 'tablet'])};
+  max-width: 52rem;
```
FrankyMartz, line 60: This dimension only applies at tablet or larger. This should be encased in a media query.

It's usually best to avoid leaving css properties active where they aren't utilized as this often leads to obscure bugs.

src/components/pages/home/FueledByCuriosity.tsx
```diff
@@ -0,0 +1,105 @@
...
+      {...props}
+    >
+      <NextImage layout="fill" alt={alt} src={src} />
+    </div>
+  );
+};
+
+const HeaderText = styled(Heading)`
+  margin-left: ${cssClamp([0, 'xl'], [15.225, 'xxl'])};
+  position: relative;
```
FrankyMartz, line 35: You should not apply position to an element unless it is explicitly required to achieve a specific css layout. While the parent component requires a relative position, none of the children outside of Image; requires this.

FrankyMartz (changes requested):
I have some concerns, with circumventing media queries through `calc/clamp`, missing images and implementing `sizes` property.

The `FueledByCuriosity` heading cuts off on iPhone 5.

Screen Shot 2021-04-26 at 2 08 05 PM

src/components/pages/home/FueledByCuriosity.tsx
FrankyMartz: The NextJS document outlines

> We recommend setting sizes when using layout="responsive" or layout="fill" and your image will not be the same width as the viewport.

https://nextjs.org/docs/api-reference/next/image#sizes

PR #550, opened 2021-04-28
task: add and clean up SVG icons

asingleton6:
Close [[B2-72](https://b2io.atlassian.net/browse/B2-72)]
- Add icons to icon folder
- Removed extra layers from each SVG (example in photo below: extra circle)

Screen Shot 2021-04-28 at 4 41 20 PM


_content/icons/community.svg
```diff
@@ -0,0 +1,12 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="50" height="48" viewBox="0 0 50 48">
```
mewdriller, line 1: 🤔 Might be worth running these through [`svgo`](https://github.com/svg/svgo) (or the web GUI [svgomg](https://jakearchibald.github.io/svgomg/)). It tried it with this one and it saved a whopping 30 bytes. 🙃

_content/icons/delivery.svg
```diff
@@ -0,0 +1,15 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="27" height="48" viewBox="0 0 27 48">
```
mewdriller, line 1: 🤔 Depending on how we intend to use these (components we set props vs image source) we may want to convert these to the equivalent React components?

mewdriller (approved):
LGTM! :last_quarter_moon_with_face: :low_brightness:

Everything looks solid left a couple thoughts on the diff. If you think they're worth handling now, go for it. 😸

tborres (commented):
In response to Drew's feedback I think that we should push this branch a little further

- [x] Run each of the svgs (including the Logo and LogoWithName) through [svgomg](https://jakearchibald.github.io/svgomg/)) (🌈 TIL) to improve their efficiency
- [x] Rather than storing as svgs, create each as a React component in `/src/components/icons` (see `LogoIcon.tsx` for an example)
- [x] Create a story or a story for each icon in Storybook so that we can easily see what each icon looks like (see the `Logo.stories.tsx` for example)

_content/icons/delivery.svg
asingleton6: Done and pushed!

_content/icons/community.svg
asingleton6: Completed!

asingleton6:
@tborres @mewdriller 
Please let me know if these updated changes are good to merge in the Next branch! :)

jest.config.js
```diff
@@ -8,8 +8,8 @@ module.exports = {
   transform: {
     '^.+\\.(ts|tsx)$': 'babel-jest',
   },
-  transformIgnorePatterns: ['[/\\\\]node_modules[/\\\\].+\\.(ts|tsx)$'],
   testPathIgnorePatterns: ['<rootDir>[/\\\\](node_modules|.next)[/\\\\]'],
+  transformIgnorePatterns: ['[/\\\\]node_modules[/\\\\].+\\.(ts|tsx)$'],
```
tborres, line 12: ❓ Did this file need to be changed? My linting is complaining about the sort order difference introduced in this change.

src/components/icons/custom-icons/Community.stories.tsx
```diff
@@ -0,0 +1,17 @@
+import { Meta, Story } from '@storybook/react';
+import React from 'react';
+
+import { CommunityIcon } from '../';
+import { SvgProps } from '../types';
+
+export default {
+  component: CommunityIcon,
+  title: 'Icons/CustomIcons/CommunityIcon',
```
tborres, line 9: 💬 For this, and the other icons, I think it would be better to just have them flatly aligned listed in the icons story level

```suggestion
  title: 'Icons',
```

src/components/icons/custom-icons/Community.stories.tsx
```diff
@@ -0,0 +1,17 @@
...
+} as Meta;
+
+export const Community: Story<SvgProps> = (args) => {
+  return <CommunityIcon {...args} />;
+};
+
+Community.args = {};
```
tborres, line 16: 💬 Since this story isn't defining any default args I don't believe that it needs to be defined at all here (I removed this line and it seemed to work). This change can probably be made for the rest of the icon stories.

```suggestion
```

tborres (commented):
Sorry to keep kicking the tires on this PR. I have a handful of questions and suggestions to consider.

- Did the `jest.config.js` file need to be changed? If it needs to be non-alpha ordered then we should put a `// eslint-disable-next-line sort-keys` so that the linter doesn't complain.
- I don't think that the `custom-icons` and `CustomIcons` hierarchical layer is needed. All of these could just slot into the `icons` directory
- Icons seems to be clipped at larger custom sizes (callout for feedback from Franky below)

@FrankyMartz it seems like most, if not all, of these icons clip when the default sizing is customized. What's causing that, and how do you think that the consumer of these icons should size them (i.e. pass in both the width and height, just a scaling property, or some other approach). See below.

image

FrankyMartz:
> Sorry to keep kicking the tires on this PR. I have a handful of questions and suggestions to consider.
> 
> * Did the `jest.config.js` file need to be changed? If it needs to be non-alpha ordered then we should put a `// eslint-disable-next-line sort-keys` so that the linter doesn't complain.
> * I don't think that the `custom-icons` and `CustomIcons` hierarchical layer is needed. All of these could just slot into the `icons` directory
> * Icons seems to be clipped at larger custom sizes (callout for feedback from Franky below)
> 
> @FrankyMartz it seems like most, if not all, of these icons clip when the default sizing is customized. What's causing that, and how do you think that the consumer of these icons should size them (i.e. pass in both the width and height, just a scaling property, or some other approach). See below.
> 
> image

Some clients will set the SVG elements to a `overflow: hidden` as a means of enforcing the `viewport` dimensions.

> Although the initial value for overflow is auto, it is overwritten in the User Agent style sheet for the <svg> element when it is not the root element of a stand-alone document,
>
> https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/overflow

Unexpected clipping is also an indicator that the combination `width`, `height` is not respecting the defined aspect ratio. We really shouldn't be defining both `width` and `height` dimensions. Ideally you only define one axis and set the other axis to `auto`. The client is much more efficient at determining the remaining dimension, and handling the half pixel decimal rounding.

_**Side Note**: Initially, my concern was you had removed required `<g>` tags. Sometimes designers use grouping as a lazy hack to trim or encase paths. If I'm going to remove nested `g` tags then I usually edit through sketch, illustrator, or a vector editor._

src/components/icons/custom-icons/Community.stories.tsx
asingleton6: Done!

src/components/icons/custom-icons/Community.stories.tsx
asingleton6: Done!

jest.config.js
asingleton6: Done!

asingleton6:
@tborres @FrankyMartz @mewdriller 
I've updated the SVGs with the ones provided directly from Bonfire. They look really good and are not clipping! 
My only question now is how we should go about optimizing the SVG.

Are we okay with using [svgomg](https://jakearchibald.github.io/svgomg/)? @FrankyMartz I know you have more experience with this. What do you think?

You can look at each SVG in the link below in case you want to check out the respective byte size. 
https://drive.google.com/drive/folders/11XbXLrN_R3IMLfxXxOpWKDoAvgmL0HeD

src/components/icons/index.ts
```diff
@@ -1,2 +1,14 @@
...
+export * from './DiscoveryIcon';
+export * from './EmbeddedIcon';
+export * from './EmpathyIcon';
+export * from './ExecutionIcon';
+export * from './FullyManagedIcon';
+export * from './SupportIcon';
+export * from './SustainabilityIcon';
+export * from './TransparencyIcon';
+export * from './BusinessIcon';
+export * from './DeveloperIcon';
```
tborres, line 14: 🔤 

```suggestion
export * from './BusinessIcon';
export * from './CommunityIcon';
export * from './DeliveryIcon';
export * from './DeveloperIcon';
export * from './DiscoveryIcon';
export * from './EmbeddedIcon';
export * from './EmpathyIcon';
export * from './ExecutionIcon';
export * from './FullyManagedIcon';
export * from './LogoIcon';
export * from './MenuIcon';
export * from './SupportIcon';
export * from './SustainabilityIcon';
export * from './TransparencyIcon';
```

tborres (approved):
Looks good 👍. Just an alpha order suggestion on the index

FrankyMartz:
> @tborres @FrankyMartz @mewdriller
> I've updated the SVGs with the ones provided directly from Bonfire. They look really good and are not clipping!
> My only question now is how we should go about optimizing the SVG.
> 
> Are we okay with using [svgomg](https://jakearchibald.github.io/svgomg/)? @FrankyMartz I know you have more experience with this. What do you think?
> 
> You can look at each SVG in the link below in case you want to check out the respective byte size.
> https://drive.google.com/drive/folders/11XbXLrN_R3IMLfxXxOpWKDoAvgmL0HeD

Hmm, I've never heard of SVGOMG. Typically I use SVGO for a quick cleanup, and will typically open a vector editor if I need to optimize beyond that — usually only happens if I need to animate or isolate portions of SVG.

FrankyMartz (approved):
👍🏼 Good work.

> Only additional side note I'd like to add is that we should be aware that applying css `height`, `width` properties will override any presentational `height`, `width` attributes on the SVG.

PR #551, opened 2021-04-30
B-12 home page content layout

asingleton6:
Close [[B2-12](https://b2io.atlassian.net/browse/B2-12)]

- Setup custom spacing between components on the home page

mobile: https://zpl.io/2jgZ9wQ
tablet: https://zpl.io/agRrNZD
desktop: https://zpl.io/V1RyNQB

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -126,6 +126,8 @@ export const Footer: FC = () => {
       as="footer"
       css={css`
         background-color: ${theme.colors.darkBlueAlt};
+        margin-top: 8.536rem;
+
         ${atMinTablet} {
           background-color: unset;
         }
--- a/src/components/HomePage/index.tsx
+++ b/src/components/HomePage/index.tsx
@@ -1,5 +1,8 @@
 import { FC } from 'react';
 import { Layout } from '~/components';
+import { spacing } from '~/theme/spacing';
+import { css } from '@emotion/react';
+import { atMinTablet, atMinXL, cssClamp } from '~/theme';
 
 import { BuiltForResults } from './BuiltForResults';
 import { DesignedForHumans } from './DesignedForHumans';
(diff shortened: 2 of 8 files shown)


src/components/HomePage/index.tsx
```diff
@@ -1,5 +1,8 @@
 import { FC } from 'react';
 import { Layout } from '~/components';
+import { spacing } from '~/theme/spacing';
```
mewdriller, line 3: ⚠️ This looks like it should be exported from `~/theme`' rather than `~/theme/spacing`. Is that the case? If not, it's probably worth refactoring  `~/theme/index.ts` to have that be the case.

src/components/HomePage/index.tsx
```diff
@@ -10,10 +13,38 @@ import { HomeHero } from './HomeHero';
...
-      <HomeHero />
-      <DesignedForHumans />
-      <BuiltForResults />
-      <FeaturedCaseStudy />
+      <HomeHero
+        css={css`
+          margin-bottom: ${cssClamp([10.5, 'mobile'], [1.5, 'desktop'])};
+        `}
+      />
+      <DesignedForHumans
```
mewdriller, line 21: 🤔 I _think_ it's unrelated to the diff but I noticed that at small sizes (320px in the screenshot) the `<DesignedForHumans />` pane's text is really jammed together compared to others.

image

Maybe something for another follow-up bug card?

mewdriller (approved):
LGTM! :loudspeaker: :loud_sound:

src/components/Footer.tsx
```diff
@@ -126,6 +126,8 @@ export const Footer: FC = () => {
       as="footer"
       css={css`
         background-color: ${theme.colors.darkBlueAlt};
+        margin-top: 8.536rem;
```
tborres, line 129: 💬 Probably makes sense to move this out of Footer and into the usage of the Footer in Layout (as we discussed)

nelsW70:
Looks great Ashley, nice work! 

It has nothing to do with this task but I noticed that "Built for Results" and "Designed for Humans" do not have consistent spacing. Just created a task for it in case anyone else notices it as well.

tborres (approved):
😎 Lookin' good

src/components/HomePage/index.tsx
asingleton6: Thanks for the feedback! I was following the structure that we have for other pages, such as the FeaturedCaseStudy component (screenshot below). Should we think about re-structuring this on other pages as well?

Screen Shot 2021-04-30 at 3 48 05 PM

src/components/HomePage/index.tsx
asingleton6: I got a card created for this!

PR #552, opened 2021-05-04
B2 13 page hero component

nelsW70:
Closes [B2-13](https://b2io.atlassian.net/browse/B2-13)

- created reusable PageHero component which receives object of images from consumer and is used at the top of Approach, Culture, Contact, Careers, and Blog pages
- created Approach page
- cropped images from Zeplin image asset with aspect ratio to match the designs as design requires 4 separate aspect ratios
- added largeDesktop bp to match design

NOTES: 

- after discussion and given the requirements of the design, it seemed best to not use the Next Image component for this component
- this is essentially a proof of concept on the approach to rendering different images 
- after discussion, I did not include this as a storybook story
- adjusted the H1 font-size in typography

<details>
<summary>mobile</summary>

Screen Shot 2021-05-06 at 3 38 18 PM

</details>

<details>
<summary>tablet</summary>

Screen Shot 2021-05-06 at 3 38 50 PM
</details>

<details>
<summary>desktop</summary>

Screen Shot 2021-05-06 at 3 39 18 PM

</details>

<details>
<summary>large desktop</summary>

Screen Shot 2021-05-06 at 3 39 46 PM

</details>


tborres (commented):
`npm run build` is failing due to a conflict with the pre-existing approach in `cms.ts`. I _think_ that we should just remove the node from `cms.ts`, but we should confirm with Drew to understand his plans for `cms.ts` before merging that change.

- [ ] rather than a `shared` directory the pictures used on particular screens should be in their own named folder (i.e. `public/approach/hero-large.jpg`)
- [ ] Move the component to `/components/PageHero.tsx` rather than nesting having a `shared`.

src/components/shared/PageHero.tsx
```diff
@@ -0,0 +1,122 @@
...
+import {
+  atMinLg,
+  atMinDesktop,
+  atMinLargeDesktop,
+  atMinTablet,
+  bp,
+  BreakpointName,
+} from '~/theme/breakpoints';
+import { spacing } from '~/theme/spacing';
+import { cssClamp } from '~/theme/util';
```
tborres, line 15: 🔧 We should update the `theme.ts` to export all of these. That would allow for everything to be imported from `~/theme`

src/components/shared/PageHero.tsx
```diff
@@ -0,0 +1,122 @@
...
+export const PageHero: FC<PageHeroProps> = ({
+  alt,
+  imgSource,
+  text,
+  ...props
+}) => {
+  return (
+    <section
+      css={css`
+        /* TODO: delete padding this once Layout component is fixed */
```
tborres, line 104: ❓ What is the expectation of the "Layout component fix"? Since this spacing doesn't seem to be related to the specifics of this component consider lifting it up to the usage (`approach/index.tsx`) at least for the time being.

src/components/shared/PageHero.tsx
```diff
@@ -0,0 +1,122 @@
...
+    left: 42rem;
+  }
+`;
+
+const ImageContainer: FC<HeroImageProps> = ({ alt, imgSource }) => {
+  const sortedImgSourcesDescending = Object.entries(imgSource).sort(
+    ([breakpointA], [breakpointB]) => {
+      const valueA = breakpointA as BreakpointName;
+      const valueB = breakpointB as BreakpointName;
+      return bp[valueB] - bp[valueA];
```
tborres, line 74: 💬 I think that the name "key" rather than "value" makes more sense here.

```suggestion
      const keyA = breakpointA as BreakpointName;
      const keyB = breakpointB as BreakpointName;
      return bp[keyB] - bp[keyA];
```

src/components/shared/PageHero.tsx
```diff
@@ -0,0 +1,122 @@
...
+  }
+`;
+
+const HeaderText = styled(Heading)`
+  bottom: 4.5rem;
+  position: relative;
+
+  ${atMinTablet} {
+    bottom: 13.75rem;
+    font-size: 6.25rem;
```
tborres, line 52: 🔧 We really shouldn't need to change the font-size for any of our typography outside of the typography file. In this case I think that the `h1` styling should be adjusted to clamp at 'tablet` rather than `lg` for both font size and letter spacing.

src/components/shared/PageHero.tsx
nelsW70: @tborres, should this be a separate task?

src/components/shared/PageHero.tsx
nelsW70: @tborres, we discussed this in standup. Once B2-82 is merged in, I'll update this to reflect those changes

src/components/shared/PageHero.tsx
nelsW70: That makes way more sense, thanks!

src/components/shared/PageHero.tsx
nelsW70: Resolved in standup

src/components/shared/PageHero.tsx
tborres: Sure

src/components/PageHero.tsx
```diff
@@ -0,0 +1,108 @@
...
+        );
+      })}
+      <img alt={alt} />
+    </Image>
+  );
+};
+
+export const PageHero: FC<PageHeroProps> = ({ alt, imgSource, text }) => {
+  return (
+    <Hero>
```
tborres, line 103: 💬 This is a situation where I think that it's easier to follow if the the root nodes styling is in place rather than being extracted to a styled component. E.g.

```
<section css={css`
  position: relative;
  ${atMinDesktop} {
    min-height: ${calculatedImageHeight};
  }
`}
```

rather than having the `Hero` variable

src/components/PageHero.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  img {
+    filter: brightness(0.85);
+    height: ${calculatedImageHeight};
+    position: absolute;
+  }
+`;
+
+const HeaderText = styled(Heading)`
+  margin-top: ${cssClamp([15, 'xs'], [20, 'tablet'])};
```
tborres, line 51: 🔧 🎨 At 375 the text appears a bit lower than it does in the comps. because the clamp calc starting at xs doesn't quite hit right.

I think that this looks a bit closer
```suggestion
  margin-top: ${cssClamp([12.5, 'smMobile'], [15, 'mobile'], [20, 'tablet'])};
```

src/components/PageHero.tsx
```diff
@@ -0,0 +1,108 @@
...
+  position: relative;
+
+  ${atMinLg} {
+    max-width: 50rem;
+  }
+
+  ${atMinDesktop} {
```
tborres, line 58: 🎨 Fun fun. The Zeplin desktop size for _this_ comp is 1440 (which we don't have as a named breakpoint yet). As this is implemented now the text doesn't come up and to the right until 1600, but the comps have it come up at 1440.

src/components/PageHero.tsx
```diff
@@ -0,0 +1,108 @@
...
+  [17.5, 'smMobile'],
+  [20.188, 'mobile'],
+  [35, 'tablet'],
+);
+
+const Image = styled.picture`
+  margin-left: calc(50% - 50vw);
+
+  ${atMinLargeDesktop} {
+    margin-left: ${cssClamp([-20, 'largeDesktop'], [-20, 'xlDesktop'])};
```
tborres, line 40: 🔧 No need for the clamp if the values aren't change. However, looking over the comps it does look like the image was too far to the left on the xlDesktop size.

```suggestion
    margin-left: ${cssClamp([-20, 'largeDesktop'], [-15, 'xlDesktop'])};
```

src/components/PageHero.tsx
```diff
@@ -0,0 +1,108 @@
...
+};
+
+type PageHeroProps = HeroImageProps & {
+  text: string;
+};
+
+const calculatedImageHeight = cssClamp(
+  [17.5, 'smMobile'],
+  [20.188, 'mobile'],
+  [35, 'tablet'],
```
tborres, line 33: ❓ Where do these numbers come from? Will this be impacted if the positioning of the text or the height of the image is changed elsewhere in this file? If so, then it would be better to pull those other numbers out as variables so that changes are consistent across the board.

tborres (commented):
Looking really good. Aside from the suggestions and questions that I added this will be ready to merge IMO.

src/components/PageHero.tsx
nelsW70: Makes sense, I toggled back and forth on that multiple times

src/components/PageHero.tsx
nelsW70: @tony, I'm pretty sure the comps have it at 1600 but could be wrong

Screen Shot 2021-05-07 at 10 48 55 AM

src/components/PageHero.tsx
nelsW70: Yeah, that's completely dumb on my part. Thanks for catching that!

src/components/PageHero.tsx
nelsW70: Whoops, thanks Tony!

src/components/PageHero.tsx
chriswhawkins: At least in this case, I think the cssClamp util might be overkill. Setting the height with a vw value (and a max-height in px or rem) likely achieves the same computed height with much less cognitive overhead. That said, if there are established patterns and/or layout factors that I'm not considering, this looks and works great!

src/components/PageHero.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  img {
+    filter: brightness(0.85);
+    height: ${calculatedImageHeight};
+    position: absolute;
+  }
+`;
+
+const HeaderText = styled(Heading)`
+  margin-top: ${cssClamp([12.5, 'smMobile'], [15, 'mobile'], [20, 'tablet'])};
```
chriswhawkins, line 52: Related to my previous comment, if the image height value is simplified, then this margin-top likely becomes easier to explicitly tie to that.

chriswhawkins (approved):
Looks really great! I left a couple of comments as "food for thought" around using cssClamp, but nothing that needs to hold up this PR. 

In talking to Nelson, it sounds like we moved away from using the Next Image component b/c the cropping is different at the different screen sizes - I'd have to really dig into those comps to understand a bit more, but it seems like we may want to look into trying to achieve the same effect visually without the need to manually create the images everywhere we need a hero.

tborres (approved):
😎 Do it!

asingleton6 (approved):

PR #556, opened 2021-05-07
B2 83 remove cms file

nelsW70:
PR:
Closes [B2-83](https://b2io.atlassian.net/browse/B2-83)

- deleted cms file and references to it in page index files
- deleted page slug files
- created culture and contact folders and index files


asingleton6 (approved):
Looks good to me!

tborres (approved):
Everything seems to be working. I'm interested to hear from @mewdriller if we should keep the `slug` files for work and blog, but we could always revive them when we need them.

🚢

nelsW70:
> Everything seems to be working. I'm interested to hear from @mewdriller if we should keep the `slug` files for work and blog, but we could always revive them when we need them.
> 
> 🚢

@tborres, should I add them back in without imports or just merge?

tborres:
Merge it!

PR #557, opened 2021-05-10
update: refactor picture, img, and HeaderText styling to give picture…

cah-nelson-foltz:
Closes [B2-88](https://b2io.atlassian.net/browse/B2-88)

- updated picture and img tags styling to give picture element same height as img 
- position absolute on picture rather than img
- updated HeaderText to use padding-top and adjusted spacing to match design


cah-nelson-foltz:
signed in with wrong github account

PR #563, opened 2021-05-12
B2 77 how we build pane

asingleton6:
Closes [B2-77](https://b2io.atlassian.net/browse/B2-77)

These components were added:
- Pictures 
- Graphics
- Titles/text
- Icon cards

Mobile: https://zpl.io/VQ607v5
Tablet: https://zpl.io/2G8EwDJ
Desktop: https://zpl.io/a7XkEpp
Extra Wide Desktop: https://zpl.io/25QE9PW


nelsW70:
This looks great @asingleton6! 

The one thing we might want to talk about with @tborres is that this approach to rendering images is a hybrid of the approach on  BuiltForResults and the PageHero. We should probably unify things where possible for consistency. 

In my opinion, this design is really using three different images in terms of aspect ratio. I'd suggest adding another image that's a square for mobile and using the approach from PageHero overall instead of the next image component. That approach skips the next image because the advantage of the next image seems to be lost once you're rendering multiple images from breakpoint to breakpoint. 

I made some changes to that approach earlier this week so it was kinda in flux when you started working on this. It does keep the picture and img elements the same size by allowing the img to grow to its parent though which isn't the case right now with this approach. 

This is all more about unifying our approach but what you're doing definitely looks great and of course works totally fine! I'm just nerding out because I stared at/wrestled with the image stuff for way too long. Just yell if you want to talk any of this over and we can get Tony's input later on!

nelsW70 (approved):

src/components/ApproachPage/HowWeBuild.tsx
```diff
@@ -0,0 +1,177 @@
+import { css } from '@emotion/react';
+import styled from '@emotion/styled';
+import NextImage from 'next/image';
+import { FC } from 'react';
+
+import { Heading, IconCard, Link } from '~/components';
+import {
+  atMinDesktop,
```
tborres, line 8: 🎨 I think that all of the usages of `atMinDesktop` in this component (and on the Approach page in general) probably should kick in at `atMinXl` instead. That's a comment that's a bit outside of this PR and we should talk about as a team.

src/components/ApproachPage/HowWeBuild.tsx
```diff
@@ -0,0 +1,177 @@
...
+          margin: 0 7.6125rem;
+        }
+      `}
+    >
+      <div
+        css={css`
+          position: relative;
+        `}
+      >
+        <picture>
```
tborres, line 81: 💬 We should see if this can leverage the semi-reusable component that Nelson has been working on.

tborres (commented):
Looking really good. A couple of miscellaneous comments at this point. I ran out of time before finishing the review.

src/components/ApproachPage/index.tsx
```diff
@@ -4,6 +4,7 @@ import { Layout, PageHero } from '~/components';
 import { atMinTablet } from '~/theme';
 import { spacing } from '~/theme/spacing';
 import { HowWeEngage } from './HowWeEngage';
+import { HowWeBuild } from './HowWeBuild';
```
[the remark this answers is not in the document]
asingleton6: Done!

src/components/ApproachPage/HowWeBuild.tsx
```diff
@@ -0,0 +1,177 @@
...
+          and share it with the world.
+        </IconCard>
+        <IconCard icon="support" heading="Support">
+          Support completes the loop, but not our relationship. During this
+          phase, a concentrated team of developers will be available to fix bugs
+          and make any necessary adjustments to the structure, flow and
+          documentation.
+        </IconCard>
+      </IconRow>
+      <Link href="" variant="CTA">
```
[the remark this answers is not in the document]
asingleton6: Done!

src/components/ApproachPage/HowWeBuild.tsx
```diff
@@ -0,0 +1,177 @@
...
+  atMinLargeDesktop,
+  atMinTablet,
+  cssClamp,
+  spacing,
+} from '~/theme';
+import { ImageProps } from '~/types';
+
+const imageDimensionsCalc = cssClamp([25.625, 'mobile'], [69.5, 'tablet']);
+
+const GradientCircleImage: FC<ImageProps> = ({ alt, src, ...props }) => {
```
tborres, line 18: 🔧 Since this component is literally called "GradientCircleImage" there's no need to pass the alt and src in. They can just be hard coded directly into this component. Essentially the component is less about re-use at this point and more about readability.

At that point this component can then just be `GradientCircleImage: FC = ()`...

tborres (approved):
Minor typo (probably from the comps), and a code adjustment suggestion.

Ready to 🚢 at your discretion.

src/components/ApproachPage/HowWeBuild.tsx
asingleton6: This has been added as a card in the backlogs.

src/components/ApproachPage/HowWeBuild.tsx
asingleton6: We agreed to update the entire Approach page once I start working on the content layout.

src/components/ApproachPage/HowWeBuild.tsx
asingleton6: done!!

PR #564, opened 2021-05-12
B2 90 culture hero page setup

nelsW70:
PR:
Closes [B2-90](https://b2io.atlassian.net/browse/B2-90)

- created Culture page using same pattern as Approach and Home
- cropped hero image with aspect ratio to match designs
- added PageHero with culture images and text
- updated max-width on hero text to align with designs


src/components/PageHero.tsx
```diff
@@ -54,7 +54,7 @@ const HeaderText = styled(Heading)`
   position: relative;
 
   ${atMinLg} {
-    max-width: 50rem;
+    max-width: 60rem;
```
tborres, line 57: ❓ Is this size change appropriate for all usages of the `PageHero`?

tborres (approved):
Looks good. Just one sanity check question.

src/components/PageHero.tsx
nelsW70: @tborres, it seems to work for Approach and Culture. I'll have to crop the images for the other pages to be sure though.

PR #566, opened 2021-05-12
Add open positions text and CTA to Culture page

mewdriller:
[B2-18](https://b2io.atlassian.net/browse/B2-18)

### DIFF
--- a/src/components/CulturePage/index.tsx
+++ b/src/components/CulturePage/index.tsx
@@ -1,21 +1,55 @@
-import React, { FC } from 'react';
-import { Layout, PageHero } from '~/components';
+import { css } from '@emotion/react';
+import type { FC } from 'react';
 
-const heroImageSource = {
-  largeDesktop: '/culture/culture-hero-xlarge.jpg',
-  tablet: '/culture/culture-hero-medium.jpg',
-  xl: '/culture/culture-hero-large.jpg',
-  xs: '/culture/culture-hero-small.jpg',
-};
+import { Heading } from '../Heading';
+import { Layout } from '../Layout';
+import { Link } from '../Link';
+import { PageHero } from '../PageHero';
+import { Text } from '../Text';
 
-export const CulturePage: FC = () => {
+export type CulturePageProps = Record<string, never>;
+
+export const CulturePage: FC<CulturePageProps> = () => {
   return (
-    <Layout>
+    <Layout title="Culture">
       <PageHero
         alt="people sitting around conference table collaborating"
-        imgSource={heroImageSource}
+        imgSource={{
+          largeDesktop: '/culture/culture-hero-xlarge.jpg',
+          tablet: '/culture/culture-hero-medium.jpg',
+          xl: '/culture/culture-hero-large.jpg',
+          xs: '/culture/culture-hero-small.jpg',
+        }}
         text="We are engineers and artisans, collaborating to make custom software people love."
       />
+      <section
+        css={css`
+          display: grid;
+          gap: 1.5rem;
+          justify-items: start;
+          margin-top: 12.5rem;
+          max-width: 52rem;
+        `}
+      >
+        <Heading as="h2" variant="h2">
+          Let&rsquo;s turn ideas into beautiful software
+        </Heading>
+        <Text as="p" variant="h3">
+          Feeling us? We&rsquo;re always looking for like-minded people with
+          similar values and development chops. Our team is tight-knit,
+          seasoned, and intentionally small&mdash;allowing everyone&rsquo;s
+          strengths to shine.
+        </Text>
+        <Link
+          css={css`
+            margin-top: 1rem;
+          `}
+          href="/culture"
+          variant="CTA"
+        >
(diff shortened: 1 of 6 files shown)


mewdriller:
FYI: I'm not entirely sure where the "See all available positions" CTA is supposed to link to, so I've left it linking back to the same page for now.

src/components/CulturePage/index.tsx
```diff
@@ -1,21 +1,55 @@
...
-        imgSource={heroImageSource}
+        imgSource={{
+          largeDesktop: '/culture/culture-hero-xlarge.jpg',
+          tablet: '/culture/culture-hero-medium.jpg',
+          xl: '/culture/culture-hero-large.jpg',
+          xs: '/culture/culture-hero-small.jpg',
+        }}
         text="We are engineers and artisans, collaborating to make custom software people love."
       />
+      <section
```
asingleton6, line 25: Will this section/pane need to be in a separate file and then imported? Kinda like how the Home and Approach page is structured.

asingleton6 (approved):

src/components/CulturePage/index.tsx
```diff
@@ -1,21 +1,55 @@
...
+        css={css`
+          display: grid;
+          gap: 1.5rem;
+          justify-items: start;
+          margin-top: 12.5rem;
+          max-width: 52rem;
+        `}
```
chriswhawkins, line 32: `display: grid` seems like overkill to me in these situations where the only change to the in-flow layout is spacing on a single axis. That said, I like how it has the potential to be a single source of truth for (or establish the default for) the vertical spacing, ie using `gap` for row spacing.

src/components/CulturePage/index.tsx
```diff
@@ -1,21 +1,55 @@
...
+        }}
         text="We are engineers and artisans, collaborating to make custom software people love."
       />
+      <section
+        css={css`
+          display: grid;
+          gap: 1.5rem;
+          justify-items: start;
+          margin-top: 12.5rem;
+          max-width: 52rem;
```
chriswhawkins, line 31: I wonder if we should start using `ch` values for these areas where width constraints are more about text placement/readability.

chriswhawkins (approved):
The text/layout looks great! I left a couple of comments that are more food-for-thought/conversational, so they shouldn't hold this up from merging.

src/components/CulturePage/index.tsx
mewdriller: Yeah, I imagine we can extract it as we determine the file is too crowded and/or there's a case for re-use somewhere else.

src/components/CulturePage/index.tsx
mewdriller: Yeah, we can definitely go with simple flow and two `margin` settings here. I ended up doing this thinking that it's a _reasonably simple_ setup but has the highest top-end for layout. Definitely open to reverting to a simpler setup for this specific use case though.

src/components/CulturePage/index.tsx
mewdriller: Yeah, I had that thought too. I guess the hesitation I had was that, in my CSS ignorance, I wasn't sure how that would play with the increased font sizes in the two children (especially as they scale up/down when the `vw` changes).

src/components/CulturePage/index.tsx
asingleton6: Cool cool! Everything else looks really good. 😃

src/components/CulturePage/index.tsx
```diff
@@ -1,21 +1,55 @@
...
+          Feeling us? We&rsquo;re always looking for like-minded people with
+          similar values and development chops. Our team is tight-knit,
+          seasoned, and intentionally small&mdash;allowing everyone&rsquo;s
+          strengths to shine.
+        </Text>
+        <Link
+          css={css`
+            margin-top: 1rem;
+          `}
+          href="/culture"
```
tborres, line 47: This should probably go to `/careers`, which hasn't been implemented yet.

src/pages/culture/index.tsx
```diff
@@ -0,0 +1,9 @@
+import type { NextPage } from 'next';
```
tborres, line 1: ❓ Is there a reason to move it from the `pages/culture.tsx` to it's own folder with an index? We've gone back and forth on this a bit and I'd like to land on a more firm rule-of-thumb. With this change the culture and approach pages follow different patterns while performing essentially the same function as a static landing page with no subdomains.

mewdriller: `pages/culture.tsx` didn't exist when I pulled the branch but `pages/culture/index.tsx` did. I'm happy to revert, but from the content on the page I was assuming there would be other pages that were the in `/culture/*` space so I used the folder + index pattern.

src/components/CulturePage/index.tsx
mewdriller: OK, I created a bug issue to track that. I'll update the card to include this information.

PR #567, opened 2021-05-13
Add Our Culture & Beliefs content to Culture page

mewdriller:
B2-89


src/components/CulturePage/OurCulture.tsx
```diff
@@ -0,0 +1,70 @@
...
+  return (
+    <section
+      css={css`
+        --repeat: auto-fit;
+        display: grid;
+        grid-template-columns: repeat(
+          var(--repeat, auto-fit),
+          minmax(15rem, 30rem)
+        );
+        gap: 6rem;
```
tborres, line 19: 🎨 If we're trying to match the comps I'm not sure if we can use the `gap` approach unless we add another wrapper around the icon card list. The mobile comp (375px) has the space between the top text and the card list as 5rem, and the space between each card as 4rem. Current implementation has it all at 6rem.

🔤 
`gap` should come before `grid-template-columns`

src/components/CulturePage/OurCulture.tsx
```diff
@@ -0,0 +1,70 @@
...
+          display: grid;
+          gap: 1.5rem;
+          grid-column: 1 / -1;
+          max-width: 52rem;
+        `}
+      >
+        <Heading as="h2" variant="h2">
+          Our Culture &amp; Beliefs
+        </Heading>
+        <Text as="p" variant="h3">
```
tborres, line 38: 💬 `Text` is `p` by default so we shouldn't need to specify it here; however, that does call out that this is a `p`, which I think some other similar panes have it as an h3. `p` is probably more appropriate, and we'll have to create a clean-up card to ensure that we're being consistent across the board.

tborres (commented):
We may want to huddle up to discuss consistency. The Our Culture is essentially the same structure as How We Engage, but the approach and some of the elements are different. I'm not sure what the "right" way should be, but they probably shouldn't be different.

I noticed at the mid-range size (I tested at 1024px) the text content layout is more narrow than the other text on the page. This is probably because it's tied to the column layout which didn't have enough room to kick into 2 columns yet.

image

mewdriller:
> We may want to huddle up to discuss consistency. The Our Culture is essentially the same structure as How We Engage, but the approach and some of the elements are different. I'm not sure what the "right" way should be, but they probably shouldn't be different.
> 
> I noticed at the mid-range size (I tested at 1024px) the text content layout is more narrow than the other text on the page. This is probably because it's tied to the column layout which didn't have enough room to kick into 2 columns yet.
> 
> image

Sounds good. I'll try to drop in on the post-coffee base2.io chat this morning.

src/components/CulturePage/OurCulture.tsx
mewdriller: Yeah. I eye-balled it and thought that it was visually consistent at an even `6rem`. There's actually a ~`9rem` horizontal gap in the columns (vs ~`6rem` on the rows) in the desktop size as well. I'll add in the exact values for all the scenarios.

src/components/CulturePage/OurCulture.tsx
mewdriller: I knew it was defaulting to `p` but I felt like adding the explicit `as="p"` made it clearer to someone coming after while not adding too much cruft. 🤷

src/components/IconCard.tsx
```diff
@@ -84,7 +82,38 @@ export const IconCard: FC<IconCardProps> = ({
...
+  );
+};
+
+export type IconCardGridProps = {
+  children: ReactElement<IconCardProps>[];
+};
+
+export const IconCardGrid: FC<IconCardGridProps> = ({ children, ...props }) => {
+  return (
+    <ul
```
tborres, line 103: At tablet size the IconGrid should be displaying in 2 columns

image

src/components/ApproachPage/index.tsx
```diff
@@ -39,6 +39,13 @@ export const ApproachPage: FC = () => {
       <QuotedImage
         alt="person standing by brick wall"
         company="Base Two"
+        css={css`
```
tborres, line 42: ⚠️ Ashley has a PR out for approach layout (#568) that may collide with this.

tborres (changes requested):
The `IconCardGrid` doesn't appear to be tiling correctly at the tablet size. Since Culture doesn't have a comp at that size you can use the Approach page comps and components as a visual guide.

nelsW70 (approved):

nelsW70:
Woohoo, so nice to see it coming together!

PR #568, opened 2021-05-13
B2 15 approach layout

asingleton6:
[B2-15](https://b2io.atlassian.net/browse/B2-15)
- Add appropriate spacing between sections
- Change the breakpoint for desktop from 1600px to 1200px

Mobile: https://zpl.io/VQ607v5
Tablet: https://zpl.io/2G8EwDJ
Desktop: https://zpl.io/a7XkEpp
Extra Wide Desktop: https://zpl.io/25QE9PW


nelsW70:
This looks great Ashley, it's awesome to see it coming together! 

The only thing I'm seeing is that the margin-bottom clamp on HowWeBuild might need another value after 'tablet' as it's too snug against QuotedImage at 1200 and above. 

Screen Shot 2021-05-14 at 8 29 40 AM

Screen Shot 2021-05-14 at 8 28 50 AM

nelsW70 (approved):

src/components/ApproachPage/index.tsx
```diff
@@ -1,8 +1,8 @@
 import { css } from '@emotion/react';
 import React, { FC } from 'react';
 import { Layout, PageHero, QuotedImage } from '~/components';
-import { atMinTablet } from '~/theme';
-import { spacing } from '~/theme/spacing';
+import { cssClamp, atMinTablet, spacing } from '~/theme';
```
tborres, line 4: 🔤 

```suggestion
import { atMinTablet, cssClamp, spacing } from '~/theme';
```

tborres (changes requested):
This is looking really good. Little tweaks:

- adjust `p` weight to 300 in the typography file
- mobile CTA seems too low
- in the 1600 comps the How We Build image should still be touching the far left, and the quote image should be touching the far right
- Starting at 1080 there should be more space between How We Build and the quote

tborres (approved):
Looks great 🌹

src/components/ApproachPage/HowWeBuild.tsx
```diff
@@ -67,14 +61,16 @@ const IconRow = styled.div`
   }
```
tborres, line 61: 🔧 ☝️ A bit above this line there is a `:nth-child(odd)` that is throwing warnings at build time because `The pseudo class ":nth-child" is potentially unsafe when doing server-side rendering. Try changing it to ":nth-of-type"`. It makes sense to make that change in this PR.

asingleton6: Done!

PR #571, opened 2021-05-18
Improve between-page font performance

mewdriller:
- Convert ttf to woff2
- Add a preload link for the font file

B2-87


mewdriller:
🤔 This cleans up the immediate issue and cuts the font size by ~60%, but we can probably get further perf gains by subsetting (only include the characters we use) the font. I was hesitant to do that in here since it seems like _some_ of the tools for subsetting can mess up the axes in variable fonts.

I think it probably makes sense to create a follow-up card to do that.

nelsW70:
@mewdriller, to get access to "font-style: italic" would we convert RoobertItalicGX.ttf to woff2 and include it in the globally below the existing font-face?

Italics are being used in the design on the Approach page in the QuotedImage component

nelsW70 (approved):

asingleton6 (approved):
This looks great! It looks great on both Mac and Windows computer too.

src/pages/_document.tsx
```diff
@@ -23,35 +23,32 @@ class CustomDocument extends Document {
...
           <NextScript />
+          <style jsx global>
+            {`
+              @font-face {
+                font-display: fallback;
+                font-family: 'Roobert';
+                font-stretch: 75% 125%;
+                font-style: oblique 0deg 20deg;
+                font-weight: 100 900;
+                src: url(/RoobertGX.woff2) format('woff2-variations');
```
tborres, line 44: ❓ Just curious, was the switch to woff2 for performance reasons or implementation?

tborres (changes requested):
This looks great. I'm only flagging this as Request changes because of Nelson's catch on the italics missing. Once that's corrected this will be ready to ship.

tborres:
Should the font loading for Storybook be changed too?

src/pages/_document.tsx
mewdriller: It's a better (more modern) format. It's ~60% smaller than the comparable `ttf`.

mewdriller:
> @mewdriller, to get access to "font-style: italic" would we convert RoobertItalicGX.ttf to woff2 and include it globally below the existing font-face in src/pages/_document.tsx?
> 
> Italics are being used in the design on the Approach page in the QuotedImage component

Good catch on the italic issue! After digging in, the variable font _only_ has an axis for weight and not slant. 😢

I ended up needing to convert the italic `ttf` to `woff2` and include that as well. Not ideal, but I think it's the best we can do for now at least.

mewdriller:
> Should the font loading for Storybook be changed too?

Missed this in the feedback. Yes, it should. I'll add a card and tackle that now.

PR #574, opened 2021-05-21
Improve a11y on icons

mewdriller:
- Use @react-aria useId to generate SSR-stable IDs for icon titles
- Add aria-labelledby and role to mark the icon SVGs
- Provide default titles for all icons, but allow for overrides

B2-86


src/components/icons/BusinessIcon.tsx
```diff
@@ -1,19 +1,28 @@
...
+  const titleId = useId(id);
+
   return (
     <svg
+      aria-labelledby={titleId}
       height={height}
+      role="img"
       viewBox="0 0 50 50"
       width={width}
       xmlns="http://www.w3.org/2000/svg"
```
chriswhawkins, line 22: I think we can lose these `xmlns` attributes if we want.

chriswhawkins (approved):
🎸🐤 LGTM! Icons test perfect in Accessibility Inspector and Axe.

src/components/icons/BusinessIcon.tsx
mewdriller: I _thought_ so, but it was in there so I wasn't sure if it there was a new best practice on SVG in React. 😊

PR #576, opened 2021-05-23
Standardize import patterns

mewdriller:
- Add eslint-plugin-import to enforce rules about import statements
- Resolve all oustanding eslint errors/warnings
- Use "import type" when only importing types
- Replace dangerous peer-index named imports with peer named imports

B2-94


nelsW70 (approved):

nelsW70:
The linter will definitely help keep those imports clean, this is awesome!

PR #578, opened 2021-06-03
B2-99 contact page hero

nelsW70:
Closes [B2-99](https://b2io.atlassian.net/browse/B2-99)

- cropped images with aspect ratio matching design for ContactPageHero
- created ContactPage folder in components and index.tsx
- deleted contact folder in src/folder because of new pattern being implemented
- added ContactPageHero
- adjusted PageHero text positioning globally to account for text that was breaking: find a happy medium

**Notes** 
- Additional polish/customization could potentially be applied to the positioning of text in relation to the image. Just wanted to get something in that works. 

<details>
<summary>mobile</summary>

Screen Shot 2021-06-03 at 4 29 30 PM

</details>

<details>
<summary>tablet</summary>

Screen Shot 2021-06-03 at 4 30 58 PM

</details>

<details>
<summary>1600</summary>

Screen Shot 2021-06-03 at 4 31 22 PM

</details>

<details>
<summary>2560</summary>

Screen Shot 2021-06-03 at 4 31 48 PM

</details>


src/components/ContactPage/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+        alt="coworkers collaborating in office"
+        css={css`
+          ${atMinXL} {
+            h1 {
+              max-width: 48rem;
+            }
+          }
+        `}
+        imgSources={[
+          ['largeDesktop', '/contact/contact-hero-xlarge.jpg'],
```
mewdriller, line 22: ⚠️ We can handle it another "optimize images" card if that's better, but there's ~50% savings possible for most of these images from a quick check in [squoosh](https://squoosh.app/).

src/components/ContactPage/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { PageHero } from '../PageHero';
+
+export const ContactPage: FC = () => {
+  return (
+    <Layout title="Contact">
+      <PageHero
+        alt="coworkers collaborating in office"
+        css={css`
+          ${atMinXL} {
+            h1 {
```
mewdriller, line 16: ⚠️ This _feels_ off to me. If we're using a `PageHero`, shouldn't it be the same across pages? Can we avoid the customization here?

src/components/PageHero.tsx
```diff
@@ -48,16 +42,12 @@ const HeaderText = styled(Heading)`
 
   ${atMinXL} {
     left: ${cssClamp(
-      [22.75, 'xl'],
+      [16.75, 'xl'],
```
mewdriller, line 45: ⚠️ I was worried this might affect other screens so I started looking around and noticed that at tablet size (`768px`) the `letter-spacing` of the text is off; according to the Approach page comps it should be `0` but it's `2px`. That changes the word breaks a bit on that copy, and probably in other places as well.

That's unrelated to the change here, but I figured it was worth mentioning.

mewdriller (approved):
LGTM! :ok: :octopus:

I had a few ⚠️ comments to resolve before merging. They could be shifted to other cards if that's a better way to keep things moving.

src/components/ContactPage/index.tsx
nelsW70: Ah, thanks Drew! I used squoosh for those just now.

src/components/ContactPage/index.tsx
nelsW70: Yeah, that's a good point. Was just trying to match the designs but found a better happy medium in the PageHero.

src/components/PageHero.tsx
nelsW70: Ah, nice catch! I adjusted the H1 letter-spacing and the positioning of the PageHero text in relation to the image accordingly

PR #579, opened 2021-06-04
B2-104 scaffold pages

nelsW70:
Closes [B2-104](https://b2io.atlassian.net/browse/B2-104)

- cropped images with aspect ratio matching design for CareersPage hero
- created BatteryXTPage, BlogPage, CareersPage, ColorSnapVisualizerPage, NetJetsPage, WorkPage basic component and returned them from their corresponding files in pages
- added Careers and Blog to navigation
- added CareersPage hero with optimized images

**Notes** 

- Seems like the CaseStudy pages are child routes of Work so I set it up that way.


src/pages/careers.tsx
```diff
@@ -0,0 +1,9 @@
+import type { NextPage } from 'next';
+
+import { CareersPage } from '~/components';
+
+const CareersNextPage: NextPage = () => {
```
mewdriller, line 5: 🤔 I think the `__NextPage` convention works well to differentiate, but I _think_ there are probably a few of the `src/pages` page components that don't follow that convention. Might be worth a quick clean-up or a follow-up card.

src/pages/work/batteryxt.tsx
```diff
@@ -0,0 +1,9 @@
+import type { NextPage } from 'next';
+
+import { BatteryXTPage } from '~/components';
+
+const BatteryXTNextPage: NextPage = () => {
```
mewdriller, line 5: 🤔 Maybe `___CaseStudy(Page|NextPage)` for these component names? It's more verbose, but it adds a little more clarity to what they are when you're looking at the folder with everything in it.

src/components/CareersPage/index.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+import { PageHero } from '../PageHero';
+
+export const CareersPage: FC = () => {
+  return (
+    <Layout title="Careers">
+      <PageHero
+        alt="person sitting at computer and smiling"
+        imgSources={[
+          ['largeDesktop', '/careers/careers-hero-xlarge.jpg'],
```
mewdriller, line 13: 🏆 Looks like these went through the optimizer already; nice work.

src/components/MobileMenu.tsx
```diff
@@ -76,6 +76,12 @@ export const MobileMenu: FC = ({ ...props }) => {
             <li>
               <Link href="/culture">Culture</Link>
             </li>
+            <li>
```
mewdriller, line 79: 🚨 On smaller mobile devices (iPhone 5/SE) this pushes the lower links off the bottom and the mobile menu doesn't scroll. We'll need to get a solution for that.

image

mewdriller (changes requested):

src/components/MobileMenu.tsx
nelsW70: There was a fairly large amount of padding on the ul so I adjusted it

src/pages/careers.tsx
nelsW70: Ok, just made those changes

src/pages/work/batteryxt.tsx
nelsW70: Sure, that makes sense

src/components/CareersPage/index.tsx
nelsW70: Thanks, squoosh is awesome!

PR #580, opened 2021-06-07
B2-22 featured case study hero

nelsW70:
Closes [B2-22](https://b2io.atlassian.net/browse/B2-22)

- created reusable CaseStudyHeroImage component
- created FeaturedCaseStudyHero component

**Notes**
- link to “View more” is pointing towards path that is currently in PR #579

**Next Steps**
- [ ] replace the preview image with unstyled asset from bonfire
- [ ] create task for the above


src/components/WorkPage/CaseStudyHeroImage.tsx
```diff
@@ -0,0 +1,14 @@
+import styled from '@emotion/styled';
+import type { FC } from 'react';
+
+import type { ImageProps } from '~/types';
+
+const HeroImage = styled.img`
+  height: 100%;
+  object-fit: cover;
+  width: 100%;
+`;
```
mewdriller, line 10: 💭 Do we have a preference established yet for inline `css={}` styles vs extracted named components with `styled()`? I tend to favor the former, but it seems like we have a smattering of both throughout the project. I think it would be good to establish a convention / heuristic to align patterns. Maybe we can chat and/or document something on this topic?

src/theme/typography.ts
```diff
@@ -89,7 +89,7 @@ export const textVariants = {
       lineHeight: 1.04,
     },
     color: colors.text,
-    fontSize: cssClamp([3.75, 'mobile'], [6.25, 'tablet']),
+    fontSize: cssClamp([3.125, 'smMobile'], [3.75, 'mobile'], [6.25, 'tablet']),
```
mewdriller, line 92: ⚠️ Is this meant to be a global change? This will change all instances of `h1` typography throughout the site.

src/components/WorkPage/FeaturedCaseStudyHero.tsx
```diff
@@ -0,0 +1,133 @@
...
+          css={css`
+            margin-bottom: ${spacing.lg};
+            margin-top: ${spacing.lg};
+            &::after {
+              margin-top: 0.25rem;
+            }
+            ${atMinTablet} {
+              margin-bottom: ${spacing.xxl2};
+              margin-top: 3rem;
+            }
```
mewdriller, line 123: 💭 Another convention question, I've tended to put newline between the `&` and media query bits of styles to visually call them out when scanning a bit more. I don't know that it's something we can `eslint`, but maybe something we should conventionalize a bit?

src/components/WorkPage/FeaturedCaseStudyHero.tsx
```diff
@@ -0,0 +1,133 @@
...
+        <ResponsiveImage
+          alt="paint a photo"
+          src="/home/featured-case-study-home-1.png"
+        />
+      </PreviewImageContainer>
+      <TextContainer>
+        <Heading
+          as="h2"
+          css={css`
+            display: none;
```
mewdriller, line 88: ⚠️ This _feels_ wrong not to show the client name as well. It's definitely like that in the design, but it just seems odd to hide the info when it could just stack vertically. It feels like having the client's name could help draw attention to various case studies if the visitor recognizes one. Thoughts?

src/components/WorkPage/FeaturedCaseStudyHero.tsx
```diff
@@ -0,0 +1,133 @@
...
+  return (
+    <div {...props}>
+      <NextImage
+        alt={alt}
+        height={706}
+        layout="responsive"
+        src={src}
+        width={400}
+      />
+    </div>
```
mewdriller, line 65: 💭 Do we need the `div` wrapper for these? Seems like just the `NextImage` should be able to work as the node?

src/components/WorkPage/FeaturedCaseStudyHero.tsx
```diff
@@ -0,0 +1,133 @@
...
+    display: block;
+    position: absolute;
+    top: 8rem;
+    width: 19.938rem;
+    & div:nth-of-type(1) {
+      border-radius: 10px;
+      box-shadow: 0 16px 21px 0 rgba(0, 0, 0, 0.53);
+    }
+  }
+`;
```
mewdriller, line 53: 💭 This structure is similar enough to the one on `HomePage` it might be worth giving the rounded corner + drop-shadow image a name and extracting it.

src/components/WorkPage/FeaturedCaseStudyHero.tsx
```diff
@@ -0,0 +1,133 @@
...
+  }
+  ${atMinDesktop} {
+    left: 50rem;
+    margin-top: unset;
+    position: absolute;
+    top: 11.5rem;
+  }
+`;
+
+const PreviewImageContainer = styled.div`
```
mewdriller, line 41: 💭 It's not called out specifically in the design, but it feels like we should probably keep the image a bit below the desktop breakpoint; maybe laptop and up?

image
image

src/components/WorkPage/FeaturedCaseStudyHero.tsx
```diff
@@ -0,0 +1,133 @@
...
+`;
+
+const TextContainer = styled.div`
+  margin-top: -3.5rem;
+  ${atMinTablet} {
+    margin-top: -6rem;
+    max-width: 35rem;
+  }
+  ${atMinDesktop} {
+    left: 50rem;
```
mewdriller, line 34: ⚠️ Maybe anchored with `right` instead of `left` here? It seems like it's more top-right than top-left from where it's placed.

src/components/WorkPage/FeaturedCaseStudyHero.tsx
```diff
@@ -0,0 +1,133 @@
...
+const PreviewImageContainer = styled.div`
+  display: none;
+  ${atMinDesktop} {
+    display: block;
+    position: absolute;
+    top: 8rem;
+    width: 19.938rem;
+    & div:nth-of-type(1) {
+      border-radius: 10px;
+      box-shadow: 0 16px 21px 0 rgba(0, 0, 0, 0.53);
```
mewdriller, line 50: 💭 Should we have a name for this shadow in the theme? Material Design has a concept of [shadows by elevation](https://material.io/design/environment/elevation.html#depicting-elevation) that's probably a good fit in terms of metaphor.

mewdriller (approved):
LGTM!

I had some feedback it might be worth discussing if we can find some time. But don't feel like you need to hold up the card for that.

src/theme/typography.ts
nelsW70: Yes, I wanted to avoid customizing a specific instance of it and thought that since 375 is the cutoff that was discussed for full support this might be a good solution

src/components/WorkPage/FeaturedCaseStudyHero.tsx
nelsW70: Ah ok, that's totally my bad and I'll adjust it

src/components/WorkPage/FeaturedCaseStudyHero.tsx
nelsW70: Yeah, it seems really wrong. I added it in and adjusted the positioning

PR #585, opened 2021-06-17
WIP: test netlify form submission

tborres:


src/components/organisms/jobs/JobForm.jsx
```diff
@@ -11,7 +11,7 @@ import {
 import { emailPattern, phonePattern } from '../../../util/regexPatterns';
 import { mediaQuery, themed } from '../../../util/style';
 
-const FILE_SIZE_LIMIT = 1000000; // 1MB
+const FILE_SIZE_LIMIT = 5000000; // 1MB
```
tborres, line 14:
```suggestion
const FILE_SIZE_LIMIT = 5000000; // 5MB
```

tborres (commented):
@mewdriller I'm going to merge this because it's better than what we have with getsimpleform (where we're just losing resumes). 

There's a couple of odd things with this current solution that we should try to iron out tomorrow.
- [ ] If the location is selected as "Other" then the custom location doesn't come over, just the word "other"
- [ ] The label for location is coming over as "Columbus, OH". So if you select Pittsburgh it says: "Columbus, OH: pittsburgh"

mewdriller:
> @mewdriller I'm going to merge this because it's better than what we have with getsimpleform (where we're just losing resumes).
> 
> There's a couple of odd things with this current solution that we should try to iron out tomorrow.
> 
> * [ ]  If the location is selected as "Other" then the custom location doesn't come over, just the word "other"
> * [ ]  The label for location is coming over as "Columbus, OH". So if you select Pittsburgh it says: "Columbus, OH: pittsburgh"

The "Other" issue is probably because we only conditionally add those fields to the from; Netlify probably needs them to be there in the SSR output so it knows how to process the field.

The "Columbus, OH" issue is probably a label/input mismatch? Not sure on that one, but I can take another spin through it.

PR #590, opened 2021-08-03
B2 95 masked images

laurapadilla:
**Kate Notes**
The images are showing changes on this branch because they have been replaced. We had originally uploaded the edited files form Zeplin (that BFR had put actual masks/effects on). I changed them to the original image files so that those effects could be implemented through CSS.

**Laura Notes**
The gradient doesn't look completely identical to the comps if you compare the images with the black backgrounds (Kate, Nelson).

We have decided to accept this as the best likeness to BFR parity. There will always be some discrepancy between design comps and implementation, and to us this was the most acceptable likeness.


src/theme/colors.ts
```diff
@@ -14,4 +14,8 @@ export const colors = {
   text: palette.offWhite,
 };
 
+//  export const gradient = {
```
k8hansen, line 17: remove

k8hansen (changes requested):
Remove the team image files that did not have changes, except for Zach's image (name change) and the noise image can stay.

Looks like the gradient could use a wee bit more tweaking to get more red happening.

laurapadilla:
I tweaked the gradient but it doesn't look 100% on images with black background. I also tweaked some of the spacing to match the Zeplin comp (adjusted the top and left margins of the team member name)

k8hansen (commented):
The images are showing changes because we had originally uploaded the edited files form Zeplin (that BFR had put actual masks/effects on). I changed them to the original image files so that those effects  could be implemented through CSS.

k8hansen (commented):
This is really close, I think it'd be better with a WEE little bit more red? If that is not possible, that's ok.

The bigger issue though is the contrast with Franky's image still. Would you be able to add a little bit of an extra dark blue gradient to the bottom of his image only? I'm meaning even doing this in photoshop and re-uploading, it just needs a bit more contrast to see his name. 

Or, if possible, increase that dark blue gradient a notch or two to make all of the images a little more contrast-y.

Spacing changes look great!

laurapadilla:
1. Added a little more red to images. 
2. Added a gradient to Frankie's image so his name stands out more

src/components/TeamMemberCard.tsx
```diff
@@ -23,19 +24,38 @@ const ResponsiveImage: FC<ImageProps> = ({ alt, src, ...props }) => {
...
+          mix-blend-mode: screen;
+        `}
       />
     </div>
   );
 };
 
+const ImageFilter = styled.div`
+  background: linear-gradient(to bottom, rgba(4, 0, 25, 0) 63%, rgba(5, 0, 30, 0.48) 92%),
+            linear-gradient(322deg, rgba(15,13,52,0.83), rgba(246,58,82,0.85));
```
k8hansen, line 38: I would change that last gradient transparency to 0.60 instead of 0.85, I'm not sure if that's what you had it at previously, but that looks the closest to me.

k8hansen (approved):
Just that one change to the gradient and then good to go!

src/components/TeamMemberCard.tsx
laurapadilla: Done!

PR #593, opened 2021-08-05
B2 117 add laura

laurapadilla:
Added myself to the new website!


k8hansen (approved):
LGTM! The aspect ratio is off, but that should be resolved in the PR that you also have open. If it is not, we will have to create a bug card. Should be ok to merge into 'next'!

PR #594, opened 2021-08-05
B2 116 fix footer safari

laurapadilla:
1. Fix alignment issue of footer links in Safari (created a new styled-component based on `Link` and adjusted the line-height)
2. Set the `font-style`  property  of the `ContactLink` styled-component to `normal` instead of italic so it matches the Zeplin comp.


laurapadilla:
Trashing this branch

PR #595, opened 2021-08-06
fix footer alignment in safari

laurapadilla:
1. Fix alignment issue of footer links in Safari (created a new styled-component based on Link and adjusted the line-height)
2. Set the font-style property of the ContactLink styled-component to normal instead of italic so it matches the Zeplin comp.

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -62,18 +62,23 @@ const Nav = styled.nav`
     }
 
     li {
-      margin-bottom: 1.25rem;
+      margin-bottom: 2rem;
 
       &:last-child {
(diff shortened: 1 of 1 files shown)


src/components/Footer.tsx
```diff
@@ -62,19 +62,24 @@ const Nav = styled.nav`
     }
 
     li {
-      margin-bottom: 1.25rem;
+      margin-bottom: 2rem;
```
k8hansen, line 65: Was this a visual change? Or to match Zeplin styles?

k8hansen (approved):
LGTM!

PR #596, opened 2021-08-09
update src path for cse study img

laurapadilla:
Update the src path for the Paint a Photo image on the Work page. 

I tried using the full image with the bottom navigation but the aspect ratio was getting lost. I tried using `object-fit` but it would just crop the image to fit inside the div, so I just ended up using the `featured-case-study-1.png` img from the `HomePage` component. I mocked it up in a codepen but wasn't able to get much farther. Open to suggestions!

### DIFF
--- a/src/components/WorkPage/FeaturedCaseStudyHero.tsx
+++ b/src/components/WorkPage/FeaturedCaseStudyHero.tsx
@@ -81,7 +81,7 @@ export const FeaturedCaseStudyHero: FC = () => {
             }
           }
         `}
-        src="/home/featured-case-study-home-1.png"
+        src={require('../HomePage/featured-case-study-1.png')}
       />
       <TextContainer>
         <Heading


src/components/WorkPage/FeaturedCaseStudyHero.tsx
```diff
@@ -81,7 +81,7 @@ export const FeaturedCaseStudyHero: FC = () => {
             }
           }
         `}
-        src="/home/featured-case-study-home-1.png"
+        src={require('../HomePage/featured-case-study-1.png')}
```
mewdriller, line 84: 💭 It feels like we might need to consider a more elegant way to `require` these images without having to struggle with where they're located relative to the usage site. Not something for now, but we could consider adding another alias like `~/images` or something?

mewdriller (approved):
LGTM! :lemon: :lock:

I made a quick note about something to consider for later, but the change-set here looks solid.

PR #597, opened 2021-08-09
B2 110 refactor team images

laurapadilla:
Added blur effect to team images


src/components/TeamMemberCard.tsx
```diff
@@ -90,17 +92,17 @@ export const TeamMemberCard: FC<TeamMemberCardProps> = ({
...
-        <Text>{position}</Text>
-      </MemberInfo>
-    </article>
+        <MemberInfo>
+          <Heading as="h3" variant="h2">
+            {name}
+          </Heading>
+          <Text>{position}</Text>
+        </MemberInfo>
+    </article>
```
mewdriller, line 106: ⚠️ It looks like the formatting might be a bit off in a few places in this file. Do you have the [prettier extension](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) installed for VS Code (or the equivalent for whatever editor you use)?

src/components/TeamMemberCard.tsx
```diff
@@ -20,6 +20,8 @@ const ResponsiveImage: FC<ImageProps> = ({ alt, src, ...props }) => {
     <div {...props}>
       <NextImage
         alt={alt}
+        blurDataURL={src}
```
mewdriller, line 23: 🚨 This looks like we're using the `src` for the blurred version of the image. If we want to leverage the blur-up functionality of the `next/image` package I believe we'll need to use the `require('../path-to-image.png')` version for `src`.

Let me know if you want to pair up on this. The `next/image` stuff is new and the documentation isn't _great_. I had to slog through figuring it all out for another project so I'm happy to spare you the struggle. 😅

mewdriller (changes requested):
I think we need to make some changes on the `src`/`blurDataUrl` bits on the `next/image` usage. Happy to help pair on that, just let me know! 😄

laurapadilla:
1. I added moved `TeamMemberCard` and `TeamCardList` files to the `Culture` folder. 
2. For some reason the grayscale property on the images was being overwritten somehow after the refactor, so I added an `!important` tag next to the property. I tried a couple of different things but this seemed to do the trick. If there's a better way of addressing this issue let me know!

mewdriller:
> 2. For some reason the grayscale property on the images was being overwritten somehow after the refactor, so I added an `!important` tag next to the property. I tried a couple of different things but this seemed to do the trick. If there's a better way of addressing this issue let me know!

I used the "Break on > attribute modifications" debugger in Chrome and apparently that's because we added the `placeholder="blur"` bit. They use a `filter: blur(20px);` value during the blur-up behavior and then set an inline `style` value of `filter: none;` after the image is ready. 

It's _really_ hacky, but this is the only way I could figure out that would add in our `grayscale(1)` directly:

image

But I think it's probably a better idea to add the `grayscale(1)` to a wrapper:

image

src/components/index.ts
```diff
@@ -19,9 +19,9 @@ export * from './MobileMenu';
 export * from './NetJetsPage';
 export * from './PageHero';
 export * from './QuotedImage';
-export * from './TeamCardList';
+export * from './CulturePage/TeamCardList';
 export * from './TeamHero';
-export * from './TeamMemberCard';
```
mewdriller, line 24: ⚠️ I think now that those components live inside `CulturePage` we *don't* want to export them from `src/components/index.ts` since we don't think anyone else should use them.

src/components/CulturePage/TeamMemberCard.tsx
```diff
@@ -91,10 +68,21 @@ export const TeamMemberCard: FC<TeamMemberCardProps> = ({
...
+          height={600}
+          layout="responsive"
+          src={src}
+          width={465}
+          css={css`
+            filter: grayscale(1) !important;
+            mix-blend-mode: screen;
+          `}
+        />
+      </ImageFilter>
```
mewdriller, line 85: ⚠️ This conflicts with the `filter: blur(20px);` value set during the blur-up behavior of `next/image`. I think this should work and isn't _too much_ additional markup:

```suggestion
      <ImageFilter>
        <ImageNoise />
        <div
          css={css`
            filter: grayscale(1);
            mix-blend-mode: screen;
          `}
        >
          <NextImage
            alt={name}
            placeholder="blur"
            height={600}
            layout="responsive"
            src={src}
            width={465}
          />
        </div>
      </ImageFilter>
```

src/components/CulturePage/TeamMemberCard.tsx
```diff
@@ -1,52 +1,37 @@
...
-        src={src}
-        width={465}
-        css={css`
-          filter: grayscale(1);
-          mix-blend-mode: screen;
-        `}
-      />
-    </div>
-  );
+  src: Exclude<NextImageProps['src'], string>;
```
mewdriller, line 14: 🔮 TypeScript magic!

src/components/CulturePage/TeamCardList.tsx
```diff
@@ -0,0 +1,130 @@
...
+  }
+`;
+
+export const TeamCardList: FC = (props) => {
+  return (
+    <TeamMasonry {...props}>
+      <TeamMemberCard
+        name="Drew Miller"
+        position="Owner & Principal Software Engineer"
+        src={require('../../../public/culture/team/dmiller.jpg')}
```
mewdriller, line 66: 💭 Eventually we probably want to co-locate these images with their usage site and/or find some better image management pattern for the project as a whole.

src/components/CulturePage/TeamCardList.tsx
```diff
@@ -0,0 +1,130 @@
...
+        padding-top: 16rem;
+      }
+    }
+  }
+
+  ${atMinDesktop} {
+    margin-left: -${spacing.xxl};
+    margin-right: -${spacing.xxl};
+    padding-bottom: ${spacing.xxl1};
+  }
```
mewdriller, line 57: 🚨 Looking over the card details, I think this is where we would want to use the `display: grid;` + `transform: translateY(var(--column-shift));` used in `FeaturedCaseStudy`.

mewdriller (changes requested):

src/components/CulturePage/TeamCardList.tsx
laurapadilla: @mewdriller For the grid, do we want to do 2 images per row, or keep it at 3? With 3 images the spacing looks a little wonky, I tried playing around with the margins but couldn't quite get it to look as nice as with just 2 images per row. Let me know if you have any suggestions or want to hop on a call!

src/components/CulturePage/TeamCardList.tsx
laurapadilla: Agreed! I left this alone for now, but I usually like to keep my component-specific images in the specific component folder but not sure if there's a better pattern out there!

src/components/CulturePage/TeamCardList.tsx
```diff
@@ -0,0 +1,125 @@
...
+  ${atMinXL} {
+    grid-template-columns: repeat(3, 1fr);
+
+    & article {
+      margin-bottom: 0;
+      margin-top: 8rem;
+      width: 100%;
+      z-index: 2;
+    }
+    > * {
```
mewdriller, line 41: 💭 Maybe a newline here to stay consistent to the formatting between the other selector groups?

```suggestion
    }
    
    > * {
```

src/components/CulturePage/TeamCardList.tsx
```diff
@@ -0,0 +1,125 @@
+import styled from '@emotion/styled';
+import type { FC } from 'react';
+
+import { atMinDesktop, atMinTablet, atMinXL, spacing } from '~/theme';
+
+import { TeamMemberCard } from './TeamMemberCard';
+
+const TeamMasonry = styled.section`
```
mewdriller, line 8: 💭 Since we're not re-using this structure anywhere, we could consider inlining it on the usage site at L#57.

src/components/CulturePage/TeamMemberCard.tsx
```diff
@@ -1,52 +1,37 @@
...
+import NextImage, { ImageProps as NextImageProps } from 'next/image';
 import type { FC } from 'react';
 
 import { atMinSm, atMinTablet, colors, spacing } from '~/theme';
-import type { ImageProps } from '~/types';
 
-import { Heading } from './Heading';
-import { Text } from './Text';
+import { Heading } from '../Heading';
+import { Text } from '../Text';
```
mewdriller, line 9: 💭 We're a little inconsistent with this throughout the codebase right now, but it should be safe (avoiding cyclical dependency loops) to import these from `~/components` now that it's extracted into the `CulturePage` subfolder.

src/components/CulturePage/TeamMemberCard.tsx
```diff
@@ -91,10 +68,24 @@ export const TeamMemberCard: FC<TeamMemberCardProps> = ({
 }) => {
   return (
     <article>
-        <ImageFilter>
-          <ImageNoise/>
-          <ResponsiveImage alt={name} src={src} />
-        </ImageFilter> 
+      <ImageFilter>
+        <ImageNoise />
```
mewdriller, line 72: 💭 Same comment as above RE potential for inlining these single-use `styled` instances as `css` usages instead.

src/components/HomePage/FeaturedCaseStudy.tsx
```diff
@@ -288,6 +288,21 @@ export const FeaturedCaseStudy: FC<FeaturedCaseStudyProps> = (props) => {
...
+          <NextImage
+            alt="bedroom with dresser and mirror"
+            placeholder="blur"
+            src={require('./featured-case-study-4.png')}
+          />
+          <NextImage
+            alt="bedroom with dresser and mirror"
+            placeholder="blur"
+            src={require('./featured-case-study-4.png')}
+          />
```
mewdriller, line 305: 🚨 Looks like we ended up with 3 additional duplicates of the `featured-case-study-4.png` image here. I think these should all be removed.

src/components/CulturePage/TeamCardList.tsx
```diff
@@ -0,0 +1,125 @@
+import styled from '@emotion/styled';
+import type { FC } from 'react';
+
+import { atMinDesktop, atMinTablet, atMinXL, spacing } from '~/theme';
+
+import { TeamMemberCard } from './TeamMemberCard';
```
mewdriller, line 6: 💭 If it makes sense, we could inline `TeamMemberCard` into this file as it's not used anywhere else. Not sure if that would make this too large to manage in this file though.

mewdriller (approved):
LGTM! :bird: :balloon:

There was one 🚨 comment that should be resolved before merging and a few 💭 to consider doing now, deferring for later, or just ignoring all together. 

Nice work!

mewdriller:
@laurapadilla: One more thing, it looks like there are some conflicts on the PR for the `TeamCardList` and `TeamMemberCard`. I think you'll need to spot-check that whatever changes were made there on `next` are properly incorporated onto this branch so we don't lose any work. 

I think those are just my formatting changes from the 9d43d81 commit but probably worth double-checking.

src/components/HomePage/FeaturedCaseStudy.tsx
laurapadilla: Oh my bad! I was testing something with the grid and forgot to remove them.

PR #598, opened 2021-08-11
Generate sitemap during build

mewdriller:
- Generate sitemap during production build in next.config.js
- Add robots.txt file
- Fix issues with font preloading
- Add security headers + CSP
- Fix linting issues in Footer, TeamCardList, and TeamMemberCard

B2-118


src/pages/_document.tsx
```diff
@@ -22,46 +22,10 @@ class CustomDocument extends Document {
...
-            rel="preload"
-          />
-          <link
-            as="font"
-            crossOrigin="anonymous"
-            href="/RoobertItalicGX.woff2"
-            rel="preload"
-          />
-        </Head>
+        <Head />
```
k8hansen, line 25: @nelsW70 do these changes look familiar to another card? I think Drew's linter is picking up some things from earlier changes that were missed on all of our branches.

public/robots.txt
```diff
@@ -0,0 +1,4 @@
+User-agent: *
+Allow: /
```
k8hansen, line 2:
```
```

public/robots.txt
```diff
@@ -0,0 +1,4 @@
+User-agent: *
+Allow: /
+
+Sitemap: https://base2.io/sitemap.xml
```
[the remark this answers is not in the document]
k8hansen: Thinking we should turn off crawling for pages that we don't have content for yet.

k8hansen (approved):
Approved with possible consideration to potentially turning off crawling for pages that are not included in this release.

public/robots.txt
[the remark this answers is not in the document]
mewdriller: 💭 Oh gotcha, we can definitely do that. I was thinking instead we would delete those pages for now and re-introduce them later. Either way should work though. I think I have a moderate preference for deleting so they're not there _at all_ rather than just hidden.

src/pages/_document.tsx
mewdriller: 💭 Oh! I understand, sorry I missed that previous discussion.

We will need to have the lock file present to ensure that when we deploy it does _exactly_ what we thought it would (so versions don't change from local to remote).

I think we can solve the different Node versions by having everyone use the latest Node v14.x (currently 14.17.4). That's the current long-term-support version of Node and what Vercel will use to build it. Keeping it consistent with the deployment platform should avoid any nasty works-on-my-machine errors due to version mismatches.

Unfortunately there's not _great_ tooling on the project to force everyone to use that version. We can specify the `engines` and `engines-strict` fields in `package.json`, but those are only enforced if someone installs our project as a dependency (which won't happen). There are also `.nvmrc` files that tell your `nvm` to swap to a 14.x version when entering the project directory. But to my knowledge, those aren't consistent between the Mac and Windows version of `nvm` and don't do anything for people that don't have `nvm` installed at all.

So I think in the end, the people working on the project need to remember to switch to `14.x` when working on it and we'll need to commit the `package-lock.json` file when it changes with cause. Thoughts?

public/robots.txt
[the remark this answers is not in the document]
k8hansen: That sounds good, I'll make a card to move them to another develop branch so that we can continue working on the next two sprints.

src/pages/_document.tsx
k8hansen: This all sounds great. I'm all about consistency, and the arguments against using a consistent version were not strong enough to have stayed in my mind.

I will update the Readme to reflect this decision.

k8hansen (approved):
Confirmed that XML sitemap is generated upon build. :shipit: 
Screen Shot 2021-08-11 at 9 47 10 AM

PR #599, opened 2021-08-11
Add basic SEO setup

mewdriller:
- Add next-seo and use in Layout
- Change Layout props to require description + title
- Add Open Graph image to public and reference properly
- Ensure www is in both sitemap and Open Graph URLs

B2-120


mewdriller:
⚠️ NOTE: We still need to add `description` values for every page that we intend to publish. I added `TODO` comments for every page but I'm not sure which ones will actually be pushed in the first release.⚠️

src/components/Layout.tsx
```diff
@@ -10,12 +10,10 @@ import { Container } from './Container';
 import { Footer } from './Footer';
 import { Header } from './Header';
 
-const Root = styled.div`
-  overflow: hidden;
```
k8hansen, line 14: Is this being replaced somewhere, or is it just no longer needed?

k8hansen (approved):
:shipit: Fantastic! It was nice to see how this was handled, and to read a little about "next-seo".

src/components/Layout.tsx
mewdriller: Yeah, it got inlined into the `div` on L#37. Since it was only used once and wasn't complex I thought that inline `css` was a bit more readable than the extracted named `styled` component.

k8hansen: Thanks for responding, understood!

PR #605, opened 2021-08-18
Limit HomeHero max-width

mewdriller:
- Limit HomeHero max-width based on the Container at mobile and above
- Add gradient overlays to left/right sides to fade lines into the bg
- Add generated safe IDs for gradients
- Generate stop nodes once rather than on every render        
- Remove extraneous motion usage on SVG
- Fix issues with TeamHero in CulturePage
- Add Eric + Javier to TeamCardList
- Remove extraneous files

B2-126


JaviAir:
Looks good to me, I was able to checkout and run it with no errors.

k8hansen (commented):
I'm trying to wrap my brain around this! 

💭 I understand why BFR gave this feedback, but I might push back a bit and say I think the dynamic element needs to extend a bit outward closer to the outmost design elements at the very least. 

The interactivity now feels cramped at lower breakpoints. Could we maybe just implement this at extra-large breakpoints? When I drop this down to mobile it feels very cramped and almost pointless. I'm including an example at a 425 mobile breakpoint where the next design element is extended to the edge of the screen. I don't see why the hero element can not do this also, as it has a gradient already built in? Just my 0.02, it's not visually displeasing as is.

Screen Shot 2021-08-18 at 1 42 31 PM

laurapadilla:
Looks good to me as well 🤠 

The only comment I have is around the unused imports in the `HomeHero.tsx` file:

Screen Shot 2021-08-18 at 3 40 30 PM

laurapadilla (approved):
Looks good to me as well 🤠

The only comment I have is around the unused imports in the `HomeHero.tsx` file:

Screen Shot 2021-08-18 at 3 40 30 PM

chriswhawkins:
Looks like this got merged already, but I somewhat agree with @k8hansen's comment above. I think it looks better at all sizes if we remove all padding:
image

mewdriller:
> Looks like this got merged already, but I somewhat agree with @k8hansen's comment above. I think it looks better at all sizes if we remove all padding:

Hmmm... definitely worth a look over. I think we probably do want to consider some tweaks to the general layout + bleed setup. It's currently more than a little clunky to get it and it feels like that should be a lot easier. Maybe we can reexamine in a follow-up card?

PR #607, opened 2021-08-18
Fix layout issues with OurTeam section

mewdriller:
- Switch offset direction for 2-column and 3-column layouts
- Use next/image with fill/cover behavior for hero
- Remove extraneous files, fix import issues
- Relocate relevant images inside CulturePage folder

B2-133


laurapadilla (approved):
Looks great to me 🤠 💫

I was trying to think of a way to achieve this effect and was toying with the `nth of type` properties so this was super helpful to see!

k8hansen (approved):
Just a secondary approval!

PR #608, opened 2021-08-19
added staff image box shadow for easy reading

JaviAir:
Card issue link: [b2-125](https://b2io.atlassian.net/browse/B2-125)

- Added box shadow to bottom of staff image.
- Created a style component to apply the box shadow style over the staff images.


src/components/CulturePage/TeamMemberCard.tsx
```diff
@@ -22,6 +22,15 @@ const ImageFilter = styled.div`
   position: relative;
 `;
 
+const ImageBoxShadow = styled.div`
+  -webkit-box-shadow: inset 0px -80px 40px -30px #05001eab;
+  box-shadow: inset 0px -80px 40px -30px #05001eab;
```
mewdriller, line 27: ⚠️ I believe `emotion` should [auto vendor-prefix](https://emotion.sh/docs/introduction#framework-agnostic) for us so we shouldn't need to manually add the `-webkit` version here.

src/components/CulturePage/TeamMemberCard.tsx
```diff
@@ -67,6 +76,7 @@ export const TeamMemberCard: FC<TeamMemberCardProps> = ({
   return (
     <article>
       <ImageFilter>
+        <ImageBoxShadow />
```
mewdriller, line 79: 💭 I had a similar tweak in my PR [here](https://github.com/b2io/base2.io/pull/607/files#diff-a39a2ed4cc61a24f4c6a63eb615e8eeb369c4157e3f559e5e697226e70f303b1R101). Comparing the two, I'm not sure at all which we'd prefer. 

`box-shadow` on the image:

image

`text-shadow` on the name:

image

Not sure which we prefer (cc: @laurapadilla, @ewaight12, @k8hansen, @chriswhawkins, @FrankyMartz) but either way it _probably_ makes sense to pull the chosen change into my PR since this file will get removed in that change-set. Sorry to have stepped on the card you were working on a bit. 😅

mewdriller: I think I have a slight preference for @JaviAir's approach (`box-shadow`) rather than mine (`text-shadow`), but I could go either way.

src/components/HomePage/HomeHero.tsx
```diff
@@ -6,11 +6,7 @@ import type { FC } from 'react';
 
 import { Heading, Text, useMouseAnimationWhileVisible } from '~/components';
 import {
-  atMinDesktop,
-  atMinLargeDesktop,
   atMinLg,
-  atMinMobile,
-  atMinXLDesktop,
```
mewdriller, line 13: 🏆 Good catch!

src/components/CulturePage/TeamMemberCard.tsx
JaviAir: Maybe a mixture of both could work. So since this file will be removed and you're pulling the chosen change into your PR, do I delete this one?

k8hansen: I think they both look great. However, I think the `box-shadow` is more in line with the general direction of the design.

mewdriller: That other PR is merged now @JaviAir, so you can move the changes onto the new file if you want. I think I agree with @k8hansen that the effect could be toned down slightly. We could also try a `linear-gradient` that is transparent all the way until the bottom edge and then provides the contrast for the white text.

JaviAir:
I rebased it off the latest next branch and put in the reduced box-shadow, how's it look now?

src/components/CulturePage/OurTeam.tsx
```diff
@@ -49,6 +49,15 @@ const TeamMemberGridItem: FC<TeamMemberGridItemProps> = ({
             width: 100%;
             z-index: 3;
           `}
+        />
+                <div
```
k8hansen, line 53: This needs help with formatting, I think a few of us are having issues with Prettier running on save. You might have to do it manually by cmd+shift+p in VScode, and run Prettier that way.

I also think that @mewdriller was hoping to find a way to do this without nexting divs, but not sure if we have resolution on that.

k8hansen (commented):
This looks fantastic!

I think we need to do one or the other in terms of treatment for legibility, so we should remove Drew's text shadow on L110 if we have decided to implement the gradient.

JaviAir:
Latest commit resolved all branch conflicts and I actually like how it looks with both styles, combined they add for greater legibility and matches with the PageHero. Could be just me though. What is the usual process for style decisions?

k8hansen (approved):
So close! This gradient looks great. 

Can you remove the text shadow on the heroes  in the _Approach_ page and the _Culture_ page? They are still showing up for me. Approved based on those changes!

Screen Shot 2021-08-23 at 2 25 31 PM

PR #609, opened 2021-08-19
B2 115 remove incomplete pages links

laurapadilla:
Remove links to incomplete pages. Please let me know if you prefer the links to be commented out rather than deleted completely.

UPDATE: Adjusted the footer layout so that everything is centered across all devices. Please let me know if these changes are okay!

🚦 **NOTE**: There were a few unused imports in the `HomeHero.tsx` file and it was causing some build errors, so I removed them. Please let me know if that's okay! 

[B2-115](https://b2io.atlassian.net/jira/software/projects/B2/boards/7?selectedIssue=B2-115)


src/components/HomePage/HomeHero.tsx
```diff
@@ -5,16 +5,7 @@ import { transparentize } from 'polished';
 import type { FC } from 'react';
 
 import { Heading, Text, useMouseAnimationWhileVisible } from '~/components';
-import {
-  atMinDesktop,
```
laurapadilla, line 9: I had to remove the unused imports since it was causing some errors.

src/components/Footer.tsx
```diff
@@ -47,7 +47,7 @@ const Nav = styled.nav`
   }
 
   ul {
-    columns: 2;
+    /* columns: 2; */
```
k8hansen, line 50: We can probably take this out if we just are removing things instead of commenting things out?

k8hansen (dismissed):
Link removal looks great!

I think we should try to center the remaining footer links at Desktop and above. It looks like they are ok on Tablet and lower.

Screen Shot 2021-08-19 at 12 14 03 PM

Screen Shot 2021-08-19 at 12 14 59 PM

laurapadilla:
Adjusted the footer layout so that it's centered on all devices. I had to remove the `flex: 1` properties and the email had a `position: absolute` and `right: 0` properties that were causing it to not wrap on desktop. Let me know if these changes are okay or if there's a better way of achieving this!

src/components/CulturePage/OurBlog.tsx
```diff
@@ -32,15 +31,6 @@ export const OurBlog: FC = (props) => {
         latest technology, teaching others is a core part of what we do. And the
         best part is, we learn more about ourselves in the process.
       </Text>
-      <Link
```
mewdriller, line 35: ⚠️ I think rather than removing just the link, we should remove the entire `OurBlog` section from the `CulturePage`. You'll probably need to grab the latest from the `next` branch to make sure you don't overwrite any changes that were previously made.

mewdriller (approved):
LGTM! :busts_in_silhouette: :bike:

I had one ⚠️ comment about the `OurBlog` section being removed but overall it looks great. Nice work!

PR #610, opened 2021-08-19
b2-122-404/500-page-finesse-and-creation

ewaight12:
feat(error-pages): 404/500 Page Finesse and Creation
- Closes [b2-122](https://b2io.atlassian.net/browse/B2-122)
- Added custom pages for 400 and 500 errors


src/theme/typography.ts
```diff
@@ -140,6 +140,9 @@ export const textVariants = {
     margin: 0,
     padding: 0,
   },
+  li: {
+    marginBottom: spacing.xs,
```
k8hansen, line 144: great catch!

ewaight12: Thanks for catching that I will fix!

src/components/Custom404Page/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+					Here are some helpful links instead:
+				</Text>
+				<ul
+					css={css`
+					list-style: none;
+					padding-left: 0;
+					
+				`}
+				>
+					<Text as="li" variant="li"><Link href="/" variant="CTA">Home</Link></Text>
```
mewdriller, line 52: ⚠️ There isn't an `li` variant for `Text` so I think these nodes can just be `<Text as="li" />`.

src/components/Custom404Page/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+					${atMinXL} {
+						margin-bottom: ${spacing.xs};
+					}
+				`}
+				>
+					We can&apos;t seem to find what you&apos;re looking for.
+				</Heading>
+				<Text as="p" variant="h3"
+					css={css`
+					${atMinXL} {
```
mewdriller, line 38: 💭 Rather than doing individual spacing between each like this at a particular breakpoint, it could be easier/cleaner to do something with `gap` for either `display: grid` or `display: flex`. Maybe something like:

image

image

src/components/Custom500Page/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { Heading, Layout } from '~/components';
+import { atMinXL, spacing } from '~/theme';
+
+export const Custom500Page: FC = () => {
+  return (
+    <Layout description="500 Error" title="Custom500ErrorPage">
+			<div
+				css={css`
+				${atMinXL} {
+					margin-left: ${spacing.marginXl};
```
mewdriller, line 13: 🔁 Similar comment as before RE ad-hoc spacing.

image

image

mewdriller: ☝️ Note: The `target` + `rel` bits will need to either wait for @laurapadilla's change to `Link`'s prop type or you'll need to replicate that change yourself.

mewdriller (changes requested):

src/theme/typography.ts
ewaight12: @k8hansen This should be fixed, thanks

k8hansen (approved):
I reviewed all of Drew's notes as well as mine. LGTM! :shipit:

PR #612, opened 2021-08-19
b2-132 added text-shadow to pageHero comp

JaviAir:
Card issue link: [b2-132](https://b2io.atlassian.net/browse/B2-132)

- Added a text-shadow style to the PageHero component
- All PageHero texts will now be easier to read when over an image.

Would you guys consider this best practice? Instead of adding the same style to 3 different pages, I just added the style once to the one shared component (PageHero).

### DIFF
--- a/src/components/PageHero.tsx
+++ b/src/components/PageHero.tsx
@@ -2,7 +2,7 @@ import { css } from '@emotion/react';
 import styled from '@emotion/styled';
 import type { FC } from 'react';
 
-import { atMinLg, atMinLargeDesktop, atMinXL, cssClamp } from '~/theme';
+import { atMinLg, atMinLargeDesktop, atMinXL, cssClamp, colors } from '~/theme';
 
 import { DynamicImage, DynamicImageProps } from './DynamicImage';
 import { Heading } from './Heading';
@@ -35,7 +35,8 @@ const Image = styled(DynamicImage)`
 const HeaderText = styled(Heading)`
   padding-top: ${cssClamp([11, 'smMobile'], [16, 'mobile'], [21.5, 'tablet'])};
   position: relative;
-
+  text-shadow: -0.25rem 0.25rem 0.375rem ${colors.darkBlue};
+  
   ${atMinLg} {
     max-width: 53rem;
   }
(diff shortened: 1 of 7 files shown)


mewdriller:
It definitely makes sense to add it to the shared component rather than in each place. I think we could consider (later) extracting the shadow definition out into the theme file and giving it a name like `shadows.md` or something so it's easier to use in other places consistently.

src/components/PageHero.tsx
```diff
@@ -2,7 +2,7 @@ import { css } from '@emotion/react';
 import styled from '@emotion/styled';
 import type { FC } from 'react';
 
-import { atMinLg, atMinLargeDesktop, atMinXL, cssClamp } from '~/theme';
+import { atMinLg, atMinLargeDesktop, atMinXL, cssClamp, colors } from '~/theme';
```
mewdriller, line 5: ⚠️ Alpha-order on this is slightly off `cssClamp` and `colors` should be reversed. The linter should be yelling at you about it, let me know if that's not happening and we can investigate the reason.

mewdriller (approved):
LGTM! :banana: :blue_book:

Nice work. I had one small ⚠️ comment about alpha-order on the named imports; should be quick to resolve.

src/components/PageHero.tsx
JaviAir: Yeah my linter is not yelling at me about import order. Even when I run the npm run lint:js command. I'm messing around with the .eslintrc file to get it to yell.

JaviAir: Figured it out, I had to add this to my eslintrc file:

    "sort-imports": ["error", {
      "ignoreCase": true,
      "ignoreDeclarationSort": true,
      "ignoreMemberSort": false,
      "memberSyntaxSortOrder": ["none", "all", "multiple", "single"],
      "allowSeparatedGroups": false
  }]

Not sure if its exactly how you guys have it set up but at least it is yelling at me about alphabetical imports.

I'll --fix all the pages with unsorted imports.

JaviAir:
Fixed all the unsorted import declarations! ✅

PR #616, opened 2021-08-20
WIP: Replace stock images

laurapadilla:
Replaced the stock images with the images Mike too of the office and with some images from the old site.


k8hansen (commented):
This image needs to be replaced with the black and white one that is in the folder. They will have a css gradient treatment on them (which is another Jira card) so we need for all of the images to be black/white. This is drew's shoulder and this image should be in that folder.
Screen Shot 2021-08-23 at 11 46 15 AM


Are there any other images of people? This heading talks about "people motivating us" and it's an empty chair! We also need to make sure the text drop shadow comes off of this once Javier's changes are in.
Screen Shot 2021-08-23 at 11 47 56 AM

This image needs to be replaced too with b/w image. It might be in the assets for bfr folder on the shared drive. 
Screen Shot 2021-08-23 at 11 49 13 AM

k8hansen (approved):
:shipit:

PR #622, opened 2021-08-30
WIP: update contact page blurb and fix autofill

laurapadilla:
1. Fix autocomplete bug on contact form.
2. Update blurb on contact page -- remove the semicolon and state from the locations list so it reads better. 

Definitely looking for some feedback on the copy for that blurb on the contact page!

### DIFF
--- a/src/components/ContactPage/ContactForm.tsx
+++ b/src/components/ContactPage/ContactForm.tsx
@@ -55,6 +55,10 @@ const FormInput = styled.input`
   &::placeholder {
     font-size: 1rem;
   }
+  &:-webkit-autofill {
+    box-shadow: 0 0 0px 1000px ${colors.darkBlue} inset;
+    transition: background-color 5000s ease-in-out 0s;
+  }
 `;
 
 const ContactAside = styled.aside`
--- a/src/components/ContactPage/ContactForm.tsx
+++ b/src/components/ContactPage/ContactForm.tsx
@@ -57,6 +57,7 @@ const FormInput = styled.input`
   }
   &:-webkit-autofill {
     box-shadow: 0 0 0px 1000px ${colors.darkBlue} inset;
+    -webkit-text-fill-color: ${colors.offWhite};
     transition: background-color 5000s ease-in-out 0s;
   }
 `;
--- a/src/components/ContactPage/ContactForm.tsx
+++ b/src/components/ContactPage/ContactForm.tsx
@@ -55,8 +55,9 @@ const FormInput = styled.input`
   &::placeholder {
     font-size: 1rem;
   }
-  &:autofill,
-  :-webkit-autofill {
+
+  &:-webkit-autofill {
+    -webkit-box-shadow: 0 0 0px 1000px ${colors.darkBlue} inset;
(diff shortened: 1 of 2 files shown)


k8hansen:
What was the auto complete bug? Is there a link to a Jira card?

k8hansen:
[Jira card with bug details.](https://b2io.atlassian.net/browse/B2-149?atlOrigin=eyJpIjoiNWE1MWVjMmU4ZjA1NDU1ODg2MDZlYWIyZjFjOTBiODYiLCJwIjoiaiJ9)

src/components/ContactPage/ContactForm.tsx
```diff
@@ -55,6 +55,10 @@ const FormInput = styled.input`
   &::placeholder {
     font-size: 1rem;
   }
+  &:-webkit-autofill {
```
asingleton6, line 58: This looks really good! Maybe for an additional browser compatibility use both `:-webkit-autofill` and `:autofill`. 
I also noticed the text color is black after autofilling on Safari. It looks like you can add this attribute below to change the color!
`-webkit-text-fill-color: ${colors.offWhite};`

https://stackoverflow.com/questions/41871618/how-to-change-safari-autofill-yellow-background-color

asingleton6 (changes requested):
Great job on this!

src/components/ContactPage/ContactForm.tsx
laurapadilla: Good call Ashley! Thank you 🙏

laurapadilla: Hmm so when I add the plain `:autofill` property it stops working...

k8hansen: emotion uses [auto vendor prefixing ](https://emotion.sh/docs/introduction#framework-agnostic)so you might not need the -webkit prefix here, [see this note from Drew](https://github.com/b2io/base2.io/pull/608#discussion_r692087230).

laurapadilla: Hmm not sure if there's something I'm missing with the emotion stuff because if I remove the `webkit` prefixes it stops working on both Chrome and Safari. The code below seems to be what makes it work on both browsers, but not sure if I'm doing something wrong:

`&:-webkit-autofill {
    box-shadow: 0 0 0px 1000px ${colors.darkBlue} inset;
    -webkit-text-fill-color: ${colors.offWhite};
    transition: background-color 5000s ease-in-out 0s;
  }`

asingleton6: We are going to use the `:-webkit-autofill` for now, since it works fine on all browsers. We will add a spike into the backlog to investigate why `:autofill` is giving us trouble and why emotion is not automating the vendor prefix. 

Great work @laurapadilla!

asingleton6 (approved):

laurapadilla:
I created a new [JIRA card](https://b2io.atlassian.net/jira/software/projects/B2/boards/7?selectedIssue=B2-159) for the autofill issue!

PR #623, opened 2021-08-30
Shift HomeHero to top of the screen

mewdriller:
B2-158

### DIFF
--- a/src/components/HomePage/HomeHero.tsx
+++ b/src/components/HomePage/HomeHero.tsx
@@ -126,16 +126,18 @@ export const HomeHero: FC = (props) => {
     );
   });
   const lineFillGradientId = useId();
-  const overlayFillGradientId = useId();
+  const sideOverlayFillGradientId = useId();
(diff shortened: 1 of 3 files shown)


k8hansen (commented):
As the `HomeHero` is moving up, it looks like it might need some extra lines at the bottom to compensate. The space in between where the subheading falls to the bottom of this hero element is larger in the mockup. I can't get an exact measurement the way this mockup is composed, but I'm measuring about `140px `between the subheading and the final line in the `HomeHero`.

Screen Shot 2021-08-31 at 11 24 00 AM
Screen Shot 2021-08-31 at 11 22 47 AM

laurapadilla (approved):
Screen Shot 2021-08-31 at 2 52 31 PM

Looks good! I agree with Kate's comments, I'll also add that in the comp it looks like the top shadow is a little bit larger than in the branch? Although it could just be an optical illusion 🔮

Screen Shot 2021-08-31 at 2 53 55 PM

PR #625, opened 2021-08-31
Fix: Adjust fontWeight + letter-spacing for CTAs

laurapadilla:
Adjusted the global font-weight for link CTAs so that it's more readable and removed the letter-spacing since it wasn't really needed.


k8hansen (approved):
I have no idea why there is a letter spacing prop on CTAs at all? I can't find this anywhere in the mockup, but I'm worried that we missed something. 

I went through a few panes that use the CTA styles and can't seem to find where this was needed ever. Approved with a little bit of aprehension.

PR #626, opened 2021-08-31
Fix: update small text color to coral

laurapadilla:
Update the _**Function. Form. ROI**_ text color to coral.


k8hansen (approved):
LGMT! 🤙 🌵

PR #627, opened 2021-08-31
b2-25-work-page

ewaight12:
feat(work-page): Work page content & layout
- Closes [#b2-25](https://b2io.atlassian.net/browse/B2-25)
- Added shared component for case study preview
- Added content for work page

Issues:
- Typescript error on line 19 of CaseStudyPreview.tsx not sure the proper way to fix feedback would be great


laurapadilla (commented):
Looks good! Only 2 callouts is that the AEP ScopeBuilder image looks a little squished, and that the left/right margins on the page as a whole are a little bigger than the comp, but I could be mistaken!

Site:
Screen Shot 2021-09-01 at 10 32 46 AM

Comp:
Screen Shot 2021-09-01 at 10 32 17 AM

AEP ScopeBuilder:
Screen Shot 2021-09-01 at 10 24 59 AM

src/components/ScopeBuilderPage/index.tsx
```diff
@@ -0,0 +1,12 @@
+import type { FC } from 'react';
+
+import { Layout } from '~/components';
+
+export const ScopeBuilderPage: FC = () => {
+  // Add description
```
k8hansen, line 6: Can delete

k8hansen:
Similar feedback to Laura's, the top/bottom margin on the header can be smaller. I would go with this `xxl2` from the comp. But this looks fantastic considering we don't really have comps for larger screen sizes. Approved with that revisitation.

Screen Shot 2021-09-01 at 3 52 48 PM

k8hansen (approved):

src/components/WorkPage/CaseStudyPreview.tsx
```diff
@@ -0,0 +1,101 @@
...
+      <NextImage
+        alt={alt}
+        height={400}
+        layout="responsive"
+        src={src}
+        width={400}
+      />
+    </div>
+  );
+};
```
mewdriller, line 28: 🚨 Looks like the build is broken due to an error in here: 

image

You can check this locally with `npm run build`.

src/components/WorkPage/CaseStudyPreview.tsx
```diff
@@ -0,0 +1,101 @@
...
+import { Link, Text } from '~/components';
+import { atMinSm, atMinTablet, colors, spacing } from '~/theme';
+
+type CaseStudyPreviewImageProps = NextImageProps & {
+  alt: string;
+  src: Exclude<NextImageProps['src'], string | StaticImageData>;
+  title: string;
+  info: string;
+  link: string;
+};
```
mewdriller, line 14: ⚠️ Looks like our linting isn't catching alpha-order on our TS types. Maybe that's something we can look into in another card to automate, but we can fix manually for now?

src/components/WorkPage/CaseStudyPreview.tsx
```diff
@@ -0,0 +1,101 @@
...
+            left: 0;
+            position: absolute;
+            top: 0.925rem;
+            width: 1.125rem;
+          }
+        }
+      `}
+      {...props}
+    >
+      <ResponsiveImage alt={alt} src={src} />
```
mewdriller, line 60: ⚠️ It looks like we're not getting any value  in encapsulating this as a local component. It's only used once and probably just adds mental overhead to the file. I think we can just inline the `next/image` here without the wrapping `div`.

src/components/WorkPage/CaseStudyPreview.tsx
```diff
@@ -0,0 +1,101 @@
...
+      <div
+        css={css`
+          margin-top: -${spacing.md};
+          padding-left: ${spacing.sm};
+          position: relative;
+          ${atMinTablet} {
```
mewdriller, line 66: 💭 Maybe a newline here between the base styles and the media query.

src/components/WorkPage/CaseStudyPreview.tsx
```diff
@@ -0,0 +1,101 @@
...
+            margin-top: -2.45rem;
+          }
+
+          ${atMinSm} {
```
mewdriller, line 70: 💭 The `sm` breakpoint is actually smaller than the `tablet` one. Maybe it would make more sense to have the `sm` styles before the `tablet` ones to mimic the mobile-first setup in how you scan the styles?

src/components/WorkPage/CaseStudyPreview.tsx
```diff
@@ -0,0 +1,101 @@
...
+            }
+          }
+        `}
+      >
+        <Link
+          href={`/work/${link}`}
+          variant="large"
+          css={css`
+            font-size: 4rem;
+          `}
```
mewdriller, line 93: ⚠️ Alpha-order on props is off here. I _think_ @JaviAir had this resolved in his updates to the `eslint` setup, but maybe not if this wasn't yelling at you.

src/components/WorkPage/index.tsx
```diff
@@ -1,14 +1,73 @@
...
+              &:nth-of-type(2n) {
+                transform: translateY(calc(var(--column-shift)));
+              }
+            }
+          }
+        `}
+      >
+        <CaseStudyPreview
+          alt={'netjets app screenshot'}
+          src={require('./images/net-jets.jpg')}
```
mewdriller, line 48: 💭 It's a small thing, but probably `netjets.jpg` is a better name. I doubt they'd ever actually look at the filename, but they don't lowercase their name with a hyphen like that.

src/components/WorkPage/CaseStudyPreview.tsx
```diff
@@ -0,0 +1,101 @@
...
+          }
+
+          ${atMinSm} {
+            & p {
+              margin-top: ${spacing.xxxs};
+              padding: 0 ${spacing.md};
+              position: relative;
+            }
+
+            & p::before {
```
mewdriller, line 77: 💭 I _think_ the `&` is extraneous here? And, if you want, you could consider nesting the `&::before` inside the `p` style block.

src/components/WorkPage/index.tsx
```diff
@@ -1,14 +1,73 @@
...
 
 export const WorkPage: FC = () => {
   // TODO: Add description for WorkPage.
   return (
     <Layout description="" title="Work">
       <FeaturedCaseStudyHero />
+      <Heading
+        as="h2"
+        variant="h1"
+        css={css`
```
mewdriller, line 17: 🔁 Same comment as before RE alpha-order.

src/components/WorkPage/index.tsx
```diff
@@ -1,14 +1,73 @@
...
+      <div
+        css={css`
+          --column-shift: ${spacing.xxl};
+          display: grid;
+          gap: ${spacing.lg};
+          grid-template-columns: minmax(350px) 1fr;
+          margin: ${spacing.xxl4} 0;
+          padding: 0;
+          position: relative;
+          ${atMinTablet} {
```
mewdriller, line 33: 🔁 Same comment as before RE newline for readability.

src/components/WorkPage/index.tsx
```diff
@@ -1,14 +1,73 @@
...
+            --column-shift: ${spacing.xxl3};
+            gap: ${spacing.xxl};
+            grid-template-columns: repeat(2, minmax(0, 1fr));
+
+            > * {
+              &:nth-of-type(2n) {
+                transform: translateY(calc(var(--column-shift)));
+              }
+            }
```
mewdriller, line 42: ⚠️ I think these should be collapsed into one `> *:nth-of-type(2n)` selector here?

src/components/WorkPage/index.tsx
```diff
@@ -1,14 +1,73 @@
...
+          margin-bottom: ${spacing.xxl};
+          margin-top: ${spacing.xxl};
+        `}
+      >
+        Unique challenges require unique solutions.
+      </Heading>
+      <div
+        css={css`
+          --column-shift: ${spacing.xxl};
+          display: grid;
```
mewdriller, line 27: 💭 Maybe a newline after vars?

src/components/WorkPage/index.tsx
```diff
@@ -1,14 +1,73 @@
...
+            }
+          }
+        `}
+      >
+        <CaseStudyPreview
+          alt={'netjets app screenshot'}
+          src={require('./images/net-jets.jpg')}
+          title="Fly Netjets"
+          info="Native application compiles to IOS & Android"
+          link="netjets"
```
mewdriller, line 51: 🔁 Same comment as before RE alpha-order.

src/components/WorkPage/index.tsx
```diff
@@ -1,14 +1,73 @@
...
+        />
+
+        <CaseStudyPreview
+          alt={'scopebuilder screenshot'}
+          src={require('./images/scopebuilder.jpg')}
+          title="AEP ScopeBuilder"
+          info="Custom web-based application"
+          link="scopebuilder"
```
mewdriller, line 59: 🔁 Same comment as before RE alpha-order.

src/components/WorkPage/index.tsx
```diff
@@ -1,14 +1,73 @@
...
+        />
+
+        <CaseStudyPreview
+          alt={'gl battery xt screenshot'}
+          src={require('./images/gl-battery.jpg')}
+          title="DNV&#8226;GL Battery XT"
+          info="Responsive web application"
+          link="batteryxt"
```
mewdriller, line 67: 🔁 Same comment as before RE alpha-order.

src/components/WorkPage/CaseStudyPreview.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+import { Link, Text } from '~/components';
+import { atMinSm, atMinTablet, colors, spacing } from '~/theme';
+
+type CaseStudyPreviewImageProps = NextImageProps & {
+  alt: string;
+  src: Exclude<NextImageProps['src'], string | StaticImageData>;
+  title: string;
+  info: string;
+  link: string;
```
mewdriller, line 13: 💭 Maybe we should rename this as `href` to be more clear what the intent of it is? And I think we should probably have consumers of the component put in the full path to the case study instead of just the last bit. That will make it more flexible if we change up where we keep those in the future; e.g.: `href="/work/batteryxt"` instead of `href="batteryxt"`.

mewdriller (changes requested):

src/components/WorkPage/CaseStudyPreview.tsx
ewaight12: Looks like this might be the best option to enforce sorting props., let me know what you think https://www.npmjs.com/package/eslint-plugin-react

PR #629, opened 2021-09-01
Fix: adjust bp for hero text

laurapadilla:
Adjust hero text breakpoint so that it starts to stack at a smaller breakpoint


k8hansen:
There is a text bumping issue happening in between breakpoints. This is pretty much unavoidable when you near some of the smaller breakpoints, but maybe there needs to be some additional padding on the `Header` to compensate?

https://youtu.be/K9GC6_MEy1s

k8hansen (approved):
Approved with some level of consideration or though to how text is bumping. Not a show stopper as it looks fine at standard screen sizes.

Looks so much closer to the comp! Great job.

PR #631, opened 2021-09-02
B2 26 tech list

k8hansen:
Co-Authored with @laurapadilla


k8hansen:
🚨  Remove code from homepage before pushing!

PR #632, opened 2021-09-08
B2 162 refactor card component

laurapadilla:
First pass at refactoring Ashley's `IconCard` component into a reusable `Card` component. 

Screen Shot 2021-09-08 at 1 17 06 PM

Screen Shot 2021-09-08 at 3 03 45 PM

Screen Shot 2021-09-08 at 3 03 21 PM

Screen Shot 2021-09-08 at 3 03 00 PM

The images above are the possible card scenarios I came across in the design comps. With the new `Card` component we should be able to recreate all of these. 

For now, I am displaying the `Card` component in the homepage just to show the different possibilities (comps above). Once it's approved I'll remove them from the home page and merge in the `chore/refactor-card-component` branch into `main`. I'll also update all the `IconCard` instances to use the new `Card` component.


src/components/Card.tsx
```diff
@@ -0,0 +1,171 @@
...
+  developer: DeveloperIcon,
+  discovery: DiscoveryIcon,
+  embedded: EmbeddedIcon,
+  empathy: EmpathyIcon,
+  execution: ExecutionIcon,
+  fullyManaged: FullyManagedIcon,
+  support: SupportIcon,
+  sustainability: SustainabilityIcon,
+  transparency: TransparencyIcon,
+};
```
mewdriller, line 68: ⚠️ This pattern of using a `string` as the value of the `Icon` prop makes it easier to use the `Card` and add an icon, as you don't have to import the icon component you want and send it to the `<Card />`. However, it also has the side-effect that any page that uses a `Card`, with or without an `icon` configured, will pay the price of importing all possible `icon` values.

In practice, that's not a _huge_ deal right now as there are only a few icons. But as a pattern, it doesn't scale well as when we add a new icon to the site you end up adding unused JavaScript to every page that uses `Card`. 

To resolve this, there are a couple refactors we can do:

1. Have `Card` accept a value of `icon` that is a `ReactNode`; e.g.: `<Card icon={<BusinessIcon />} />`.
2. Have card setup the `IconList` using `next/dynamic` to enable bundle splitting so that pages only get the icons they use.

src/components/Card.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+export type CardProps<E extends ElementType = ElementType> = {
+  as?: E;
+  icon?: Icon;
+  largeHeading?: string;
+  heading?: string;
+  link?: string;
+  subheading?: string;
+  text?: string;
+};
```
mewdriller, line 52: 💭 Are there any usage restrictions on combination of props that we want to enforce? Is it OK if someone passes `largeHeading`, `heading`, and `subheading`? If not, we can use a union in TypeScript to tell the compiler what combinations of props are valid.

src/components/HomePage/index.tsx
```diff
@@ -1,8 +1,9 @@
 import { css } from '@emotion/react';
 import type { FC } from 'react';
+import React from 'react';
```
mewdriller, line 3: 🔁 Same comment RE global `React`.

src/components/HomePage/index.tsx
```diff
@@ -54,6 +55,47 @@ export const HomePage: FC<HomePageProps> = () => {
           }
         `}
       />
+      <CardGrid>
+        <Card
+          css={css`
+            &:before {
```
mewdriller, line 61: ⚠️ @chriswhawkins might correct me here, but I believe this should be `&::before` to be _more_ correct? Either works but it's my understanding that when we're trying to create a pseudo `before` we use `::` rathe than `:`.

src/components/HomePage/index.tsx
```diff
@@ -54,6 +55,47 @@ export const HomePage: FC<HomePageProps> = () => {
...
+              background-color: ${colors.lightBlue};
+              content: '';
+              display: block;
+              height: 0.25rem;
+              margin-bottom: 1rem;
+              width: 2.5rem;
+            }
+          `}
+          largeHeading="Millions"
+          heading="of users worldwide"
```
mewdriller, line 71: ⚠️ Looks like the alpha-order on props linting isn't working properly? We should take another look at that.

src/components/HomePage/index.tsx
```diff
@@ -54,6 +55,47 @@ export const HomePage: FC<HomePageProps> = () => {
...
+        <Card
+          subheading="Joining a team — blog post"
+          text="This is some text check me out. This is some text check me out. This is some text check me out. This is some text check me out. This is some text check me out. "
+        >
+          <Link
+            css={css`
+              font-size: 2rem;
+            `}
+            color="coral"
+            href="#"
```
mewdriller, line 92: ⚠️ Similar to the placeholder text, we'll want to target the eventual path of the page here.

src/components/HomePage/index.tsx
```diff
@@ -54,6 +55,47 @@ export const HomePage: FC<HomePageProps> = () => {
...
+        />
+        <Card
+          subheading="Joining a team — blog post"
+          text="This is some text check me out. This is some text check me out. This is some text check me out. This is some text check me out. This is some text check me out. "
+        >
+          <Link
+            css={css`
+              font-size: 2rem;
+            `}
+            color="coral"
```
mewdriller, line 91: ⚠️ If we're specifying `variant="CTA"`, I _think_ we shouldn't have to customize `color` and `font-size`? Or is this a one-off CTA that we want to keep as a one-off?

Maybe it should be represented as a `CardAction` or something like that if it's the "when you have a CTA in a Card" use case?

src/components/Card.tsx
```diff
@@ -0,0 +1,171 @@
...
+  | 'delivery'
+  | 'developer'
+  | 'discovery'
+  | 'embedded'
+  | 'empathy'
+  | 'execution'
+  | 'fullyManaged'
+  | 'support'
+  | 'sustainability'
+  | 'transparency';
```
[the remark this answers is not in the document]
laurapadilla: hehe credit goes to Ashley on this one — I modeled it after her work on the `IconCard` 😃

src/components/Card.tsx
laurapadilla: Let's chat about this one!

src/components/Card.tsx
```diff
@@ -0,0 +1,171 @@
...
+          {text}
+        </Text>
+      )}
+      {children}
+    </Component>
+  );
+};
+
+export type CardGridProps = {
+  children: ReactElement<CardProps>[];
```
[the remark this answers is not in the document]
laurapadilla: Credit to Ashley on this one as well since this was in her `IconCard` component!

src/components/HomePage/index.tsx
```diff
@@ -54,6 +55,47 @@ export const HomePage: FC<HomePageProps> = () => {
...
+          subheading="Business"
+          text="This is some text check me out. This is some text check me out. This is some text check me out. This is some text check me out. "
+        />
+        <Card
+          heading="The solution"
+          text="This is some text check me out. This is some text check me out. This is some text check me out. This is some text check me out. "
+        />
+        <Card
+          subheading="Joining a team — blog post"
+          text="This is some text check me out. This is some text check me out. This is some text check me out. This is some text check me out. This is some text check me out. "
```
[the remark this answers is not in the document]
laurapadilla: Hehe absolutely, the plan is to remove all that code from the home page and then just merge in the refactor. This was just for show 😄

src/components/HomePage/index.tsx
laurapadilla: Yep! Once it's actually implemented in a real scenario it should have the actual path.

src/components/icons/DashIcon.tsx
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <svg
+      aria-labelledby={titleId}
+      height={height}
+      role="img"
+      viewBox="0 0 50 50"
+      width={width}
+      {...props}
+    >
+      <title id={titleId}>{title}</title>
```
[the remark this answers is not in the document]
laurapadilla: Ashley strikes again 💪 🙃

src/components/HomePage/index.tsx
laurapadilla: You're totally right! My bad. Maybe we can discuss this one a bit more.

src/components/HomePage/index.tsx
laurapadilla: that's what I thought but wasn't sure!

src/components/HomePage/index.tsx
laurapadilla: so weird that it didn't yell at me for it

src/components/icons/DashIcon.tsx
```diff
@@ -0,0 +1,28 @@
+import { useId } from '@react-aria/utils';
+import type { FC } from 'react';
+
+import { SvgProps } from './types';
+
+export const DashIcon: FC<SvgProps> = ({
+  height = '100%',
+  id,
+  title = 'Dash',
+  width = '100%',
```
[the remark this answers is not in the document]
laurapadilla: I meant to remove this since I ended up using `::before` but I wanted to chat more about icons and how we treat them on the site

src/components/ApproachPage/HowWeEngage.tsx
```diff
@@ -1,10 +1,10 @@
 import { css } from '@emotion/react';
 import type { FC } from 'react';
 
-import { Heading, IconCard, Link } from '~/components';
+import { Heading, Link } from '~/components';
 import { atMinTablet, atMinXL, spacing } from '~/theme';
 
-import { IconCardGrid } from '../IconCard';
+import { CardGrid, IconCard } from '../Card';
```
mewdriller, line 8: ⚠️ We should grab `CardGrid` and `IconCard` from the `~/components` import to be more concise.

```suggestion
```

src/components/ApproachPage/HowWeEngage.tsx
```diff
@@ -1,10 +1,10 @@
 import { css } from '@emotion/react';
 import type { FC } from 'react';
 
-import { Heading, IconCard, Link } from '~/components';
+import { Heading, Link } from '~/components';
```
mewdriller, line 4: 👇 We should grab `CardGrid` and `IconCard` here.

```suggestion
import { CardGrid, Heading, IconCard, Link } from '~/components';
```

src/components/CulturePage/OurCulture.tsx
```diff
@@ -3,8 +3,8 @@ import type { FC } from 'react';
 
 import { atMinXL, spacing } from '~/theme';
 
+import { CardGrid, IconCard } from '../Card';
 import { Heading } from '../Heading';
-import { IconCard, IconCardGrid } from '../IconCard';
 import { Text } from '../Text';
```
mewdriller, line 8: 🔁 Same comment as earlier, we should grab these from `'~/components'` rather than `'../Card'`; e.g.:

```
import { CardGrid, Heading, IconCard, Text } from '~/components';
```

src/components/icons/BusinessIcon.tsx
```diff
@@ -4,10 +4,10 @@ import type { FC } from 'react';
 import type { SvgProps } from './types';
 
 export const BusinessIcon: FC<SvgProps> = ({
-  height = '100%',
+  height = '48px',
   id,
   title = 'Business',
-  width = '100%',
+  width = '48px',
```
mewdriller, line 10: 💭 I think just `48` works here too for pixel values.

mewdriller (approved):
LGTM! :headphones: :hamburger:

I had a couple ⚠️ comments that should be addressed first but things look solid overall.

PR #634, opened 2021-09-14
B2 30 case study nav component

laurapadilla:
Zeplin comp for reference:
Screen Shot 2021-09-14 at 11 16 14 AM

📌 `CaseStudyNav` component is at the bottom of the homepage. I will remove once PR is approved!

Looking for feedback on this component since I feel like I over-engineered some things to get it looking like the designs 😬

📓 **A couple of notes:**
For the image overlay -- I had a little trouble getting it to work on both desktop and mobile so I just adjusted the image opacity which I think works better for readability, but open to feedback on recreating this!

On mobile, I positioned the text directly below the image since I had a bit of trouble getting this to work, but I do think it's better for readability.

### DIFF
--- a/src/components/WorkPage/CaseStudyNav.tsx
+++ b/src/components/WorkPage/CaseStudyNav.tsx
@@ -1,4 +1,5 @@
 import css from '@emotion/css';
+import styled from '@emotion/styled';
 import NextImage, { ImageProps as NextImageProps } from 'next/image';
 import { Children, cloneElement, FC, ReactElement } from 'react';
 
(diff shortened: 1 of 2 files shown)


ewaight12:
I think this looks really good @laurapadilla it's spot on with the mock. For the text placement I think it looks fine below the image but I did something similar with the text overlaying on the /work page if that helps at all to look at. I don't know if that would work being that there is an opacity on the image.

src/components/WorkPage/CaseStudyNav.tsx
```diff
@@ -0,0 +1,111 @@
...
+export const CaseStudy: FC<CaseStudyProps> = ({
+  client,
+  href,
+  caseStudy,
+  alt,
+  src,
+}) => {
+  return (
+    <div
+      css={css`
```
k8hansen, line 25: putting the css right in the component works good for really minor changes, but I think for this many style edits it would be worth pulling out into a styled component. 

@mewdriller might have better reasoning, but that's something I've noticed in other projects. Usually when the styles get complicated they are pulled out of being inline.

src/components/WorkPage/CaseStudyNav.tsx
```diff
@@ -0,0 +1,111 @@
...
+    <section
+      css={css`
+        align-items: center;
+        display: flex;
+        flex-direction: column;
+
+        margin-top: ${spacing.xxl2};
+        object-fit: contain;
+
+        ${atMinXL} {
```
k8hansen, line 97: Same here, anytime a breakpoint or many styles are needed, I would pull that `section` out into a separate styled component.

k8hansen (commented):
This looks absolutely fantastic! The changes you made are great and it's so close to the mockup. 

One minor change, the space between the company and the project is a little too far, they seem to be closer together in the comp. 

There is a vertical spacer in zeplin, but it looks like it's not flush with the text, so I'm not sure if this screenshot is helpful. Screenshot attached.

🚨  Once that is changed, make sure to pull out the component from the home page. That can probably get approval for this component.

💭  This is a question for someone more senior: We will need to refactor this component to hold all of the info used in a json file, and then populate the info programatically.  It might be confusing to go through each example and make sure the left/right components are correct. It would make more sense to hold each work sample in an array so that if for example if position 0 is being displayed, navigation shows position 6 and 1. 

This would be the best way to do this for scaleability. However, if that's overthinking it, then we can just hardcode each page. I think someone else should make that call though.

Screen Shot 2021-09-29 at 11 43 03 AM

Screen Shot 2021-09-29 at 11 41 28 AM

**For approval:**

- [ ] Fix padding
- [ ] Remove from homepage

**Next steps:**

- [ ] Refactor hold programatic content

laurapadilla:
Component was removed from home page!

For the padding, I think it should remain the same between mobile/desktop since the spacing between `project name` and `explore case study link` doesn't change between devices

PR #640, opened 2021-10-08
B2 27 horizontal scroll carousel

laurapadilla:
[Desktop Comp
](https://app.zeplin.io/project/5fff21e06012bf1d470820c3/screen/5fff5937f5f0a522bee90739)

[Mobile Comp](https://app.zeplin.io/project/5fff21e06012bf1d470820c3/screen/600b2fe270e88f11679a7c73)

Instead of doing a carousel, we are stacking the images on both desktop and mobile.

### DIFF
--- a/src/components/Card.tsx
+++ b/src/components/Card.tsx
@@ -139,12 +139,22 @@ export const ImageCard: FC<ImageCardProps> = ({ heading, src, text }) => {
           align-items: center;
           background: ${colors.offWhite};
           display: flex;
+          height: 37.5rem;
           justify-content: center;
-          margin: 0 0 1rem 0;
-          padding: ${spacing.xxl};
+          margin: 0;
+          padding: 2rem 6rem;
+          width: 100%;
         `}
       >
-        <NextImage src={src} />
+        <div
+          css={css`
+            height: 100%;
+            position: relative;
+            width: 100%;
+          `}
+        >
+          <NextImage objectFit="contain" layout="fill" src={src} />
+        </div>
       </figure>
       <Heading as="h3" color="coral" variant="h3">
         {heading}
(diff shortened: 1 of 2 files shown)


k8hansen (changes requested):
I think this looks ok on mobile, but really doesn't look polished on desktop. I think the grey really detracts from the images and is distracting and incomplete. There is a big issue with the scroll being so long, as there will be other panes here. This section should not take up 3x more space than what we have been using for a standard pane.

This is totally not a critique on the development, its really hard to develop something with no design, this is a great first step. I think some things that we can do to absolutely to improve the information on the page and the design.

- [ ] replace the second and third photo with actual photos of the application - they should be reflective of the caption or we need to redo the captions. Either way, we need pics of the actual app. The app is online here: https://www.sherwin-williams.com/visualizer#/active
- [ ] Remove the gray box
- [ ] Treat the images some way that will make them similar to all of the other images on the site
- [ ] Use an H2 on the red caption text so that it is not dwarfed by the image

I don't think the white square makes any sense to be included if we are stacking the image now that I see it all together. It's my impression that was serving as a container for the entire carousel.

How all of that comes about is to be determined and might take a little design work or collaboration. It might be best to mock something up first before you throw any more dev hours into this.

 Initial thoughts on images are: 

- Images are all duotone or have a filter we have used before until hovered over
- Images are much smaller and enlarge into a preview window when clicked (this is kind of old fashioned)

Might be worth talking to Chris about this too, I believe he worked on the project and would have some insight into what we need to showcase, in addition to likely some design feedback.

k8hansen:
I just took another look at this, several years later.

The white background in the design is really weird and is a step down from the stacked samples that we have now. Lets leave this as is, close the card, and if we ever feel the need to showcase our portfolio in a carousel, we can do so.

PR #642, opened 2021-10-11
B2 164 curiosity circle animation

laurapadilla:
Add animation to `GradientCircle` 

UPDATE: Added drew's implementation of the animation. The only thing left is removing the animation on mobile.  I tried `useMedia` to conditionally render the circle but it was messing up the animation on mobile. I also thought something like this would be useful but wasn't sure how to adjust our code: https://codesandbox.io/s/framer-motion-matchmedia-ux6uk?file=/src/App.tsx

Suggestions are welcome!

### DIFF
--- a/src/components/GradientCircle.tsx
+++ b/src/components/GradientCircle.tsx
@@ -66,6 +66,7 @@ export const GradientCircle: FC = (props) => {
       height="100"
       initial="hidden"
       ref={ref}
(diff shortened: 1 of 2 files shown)


k8hansen:
Pausing this PR, as it is deprioritized for the 2024 SEO + Client Story work.

PR #655, opened 2022-08-23
update approach and culture hero images

k8hansen:
Card is [here](https://b2io.atlassian.net/browse/B2-186?atlOrigin=eyJpIjoiNmUzMzNiZTc4MGIxNGVmYzg3YWFhOTdkYmZkNzVlNmUiLCJwIjoiaiJ9).

Updated some images showing teammates that are no longer with B2. I don't love these images, but they are a little more honest while we eventually get some better photos.

Also snuck in an update on Javier's location.

**To review:**

Check out the Culture, Approach, and Contact hero pages and alt text.


src/components/ContactPage/index.tsx
```diff
@@ -11,7 +11,7 @@ export const ContactPage: FC = () => {
   return (
     <Layout description="Like what you see? Let's get started." title="Contact">
       <PageHero
-        alt="lighting fixture mounted on a brick wall"
+        alt="group of people walking toward you."
```
tborres, line 14:
```suggestion
        alt="group of people walking toward you"
```

src/components/CulturePage/index.tsx
```diff
@@ -15,7 +15,7 @@ export const CulturePage: FC = () => {
       title="Culture"
     >
       <PageHero
-        alt="people standing and collaborating in an office"
+        alt="members of team standing in front of a comically large shoe."
```
tborres, line 18:
```suggestion
        alt="members of team standing in front of a comically large shoe"
```

tborres (commented):
Thanks for digging into these portions of the site that we've clearly been neglecting!

A couple questions / suggestions
- [x] Should `public/culture/culture-hero-large 2.jpg` have been added? It doesn't appear to be referenced anywhere. Looks like a copy pasta.
- [ ] `public/culture/culture-hero-small.jpg` is a much bigger image than the one that it is replacing. It can probably be optimized.
- [ ] The following files don't appear to be referenced, and were probably accidentally add
    - [ ] `public/culture/culture-hero-xlarge 2.jpg` 
    - [ ] `public/culture/culture-hero-large 2.jpg`

k8hansen:
Thanks for reviewing.

Image handling on the whole site needs to be organized and considered. We should be using the NextJS image component to optimize image sizes automatically.

I'm manually doing it here in the interest of time, but I made a [refactor card](https://b2io.atlassian.net/jira/software/projects/B2/boards/7) to address this.

PR #656, opened 2022-09-07
Added contrast filter

k8hansen:
Added a contrast filter so that the Page Hero text has more contrast and is easier to read.


tborres (approved):
I'm fine with this, but I wouldn't mind another opinion. Both the original and the new contrast have the same accessibility score, and I'm not sure that the new contrast helps me read the text or not.

**New**
image

**Original**
image

PR #657, opened 2022-09-07
Feature/case studies  image grid

k8hansen:
The goal of this card was to replace the placeholder latin data grid with some images, in the interest of launching the Case Studies feature that has been stalled for some time.

Despite the[ Jira card title,](https://b2io.atlassian.net/browse/B2-187) the ultimate solution was not a grid, as the images were not scannable or readable. After consulting with some colleagues, this was the direction we thought would be best to go in while staying away from a light-box.

I did try to add some of the ColorBlock design elements in the back, but it just looked too busy and separate from the _Results_ section.

The images are just in a column on smaller screens, and slightly staggered on larger screens.

I welcome any suggestions, both in design and code! I assume there is a better way to handle these breakpoints, but my brain is stuck on NJ dotcom and that's how we handled it there!

**NOTE:** this is a branch off of the Case Study feature branch.

***

https://user-images.githubusercontent.com/13206464/188969114-8f2fb452-6d94-4e5f-b20a-911c8b62711b.mp4

https://user-images.githubusercontent.com/13206464/188969126-4a72cc29-eed7-4893-adf5-b25e763f372b.mp4


src/components/ColorSnapVisualizerPage/ColorSnapImagePanel.tsx
```diff
@@ -0,0 +1,113 @@
...
+        padding: 1rem 0;
+        display: inline-block;
+        position: relative;
+
+        ${atMinLg} {
+          padding: 3rem 0;
+          width: 85%;
+
+          &:nth-of-type(odd) {
+            right: calc(40% - 50vw) !important;
```
k8hansen, line 36: Yes. I used !important. NextJS image seems to be forcing some sort of display properties that I spent hours and hours trying to uncover. This was used elsewhere that had been green-lighted so I did it.

zachcd (changes requested):
This looks good except for on ultrawide screens the -50vw rule makes the image go way out there.  After that's changed, this looks great!

src/components/ColorSnapVisualizerPage/index.tsx
```diff
@@ -112,6 +106,7 @@ export const ColorSnapVisualizerPage: FC = () => {
         alt="screenshot"
       />
       <CaseStudyResultsPane {...CaseStudyResultsProps} />
+      <ColorSnapImagePanel alt={''} src={''} />
```
tborres, line 109: ❓ empty string for alt and src?

src/components/ColorSnapVisualizerPage/ColorSnapImagePanel.tsx
k8hansen: @tborres yes, I actually just came back to say this. It was being used in the Our Team section that Drew made, so I thought it might have been A Thing.

k8hansen: > On my desktop, when I fullscreen this on a 3440X1440, I see the images all the way to the right, like this: image
> 
> I think that the -50vw is too much when the view width is very wide

Well that's new! Thanks @zachcd, great catch.

src/components/ColorSnapVisualizerPage/index.tsx
k8hansen: Making these types optional got rid of this error. That seems not quite right to me, am I missing something?

k8hansen:
So in this update I addressed the changes that were out of scope of his PR but needed to be done:

- [x] Added CTA to featured case study on home page
- [x] Added Colorsnap to the nav panel at the bottom. I also rearranged it so the left panel of each page will bring you to the "previous" case study, and the right serves as "next". You should be able to cycle through and see everything this way.

I have no idea why I was trying to align these using positioning, so I switched to flexbox. Everything should look better at larger screens.

Lastly, I fixed the registration mark error, this was likely something I made when I was trying to make it smaller and fix a similar issue on medium screens.

sludemann (approved):
Looking fantastic on my end. Image changes and content changes all look good.  

Case Study style wise, and not really relevant to this PR so don't fix it here but tablet sizes the description text glues a little uncomfortably close to the left edge of the screen:
Looking fantastic on my end. Image changes and content changes all look good.  
image

PR #685, opened 2023-09-03
remove Eric & add Seattle

tborres:


mewdriller (changes requested):
Unfortunately it looks like we'll have to upgrade to Node 18.x before we can continue deployments on Vercel:

```
Error: Node.js Version "14.x" is discontinued and must be upgraded. Please set Node.js Version to 18.x in your Project Settings to use Node.js 18.
```

PR #686, opened 2023-09-07
Fix: emotion issue

nelsW70:
- pinned Node version via Volta
- updated eslintrc
- updated tsconfig
- updated imports of emotion


.eslintrc
```diff
@@ -66,6 +66,7 @@
       }
     ],
     "react/display-name": 0,
+    "react/no-unknown-property": ["error", { "ignore": ["css"] }],
```
tborres, line 69: ❓ Was this per emotion's documentation?

tsconfig.json
```diff
@@ -17,8 +18,8 @@
     "skipLibCheck": true,
     "strict": true,
     "target": "es5",
-    "incremental": true,
-    "types": ["@emotion/core"]
+    "types": ["@emotion/react/types/css-prop"],
+    "incremental": true
```
tborres, line 22: ❓ Does this need to come at the end, or can these be placed in alpha-order?

tborres (approved):
Everything seems to be running correctly. Just a couple of sanity-check questions. Thanks for handling this.

.eslintrc
nelsW70: @tborres, was looking at that before you commented. I don't think it's necessary with the update to the tsconfig. 

Just removed it and am not seeing any issues

tsconfig.json
nelsW70: @tborres, there don't seem to be any issues alpha-ordering them

Issue #690, opened 2023-09-08
pre-fetch-src console error


nelsW70:
![Image](https://user-images.githubusercontent.com/17680907/266676691-52896a29-583e-402e-b949-e5246f464b8b.png)

JaviAir:
PR: https://github.com/b2io/base2.io/pull/730

PR #730, opened 2023-11-28
Cleaned up some console errors.

JaviAir:
This PR tackles cards:
https://github.com/b2io/base2.io/issues/690
https://github.com/b2io/base2.io/issues/721

🩹  workaround for old nextjs es lint bug,
https://github.com/vercel/next.js/issues/28971
Bug found in old version of next / eslint dependecies, disabling the rule is a temporary workaround. Blocked from building the app without the rule off. True fix is by updating dependencies which is its own task/pr/card. 

➖  deprecated prefetch src csp,
image
The Prefetch src CSP rule is deprecated and unrecognized by chromium/Firefox. Safe to delete.

🩹  workaround preload attribute on italic font link
App struggled to preload an italic font link. Workaround was to remove the preload attribute. No discernible performance impact without the preload.

* Note: If viewing the vercel live preview, there will be a script-src-elm csp console error but that is only specific to the preview. To get a more accurate reading, you would have to `npm run build` and `npm run start `the app locally. 

Console should be squeaky clean! 👍 
...for now 👀

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -29,6 +29,7 @@
     }
   },
   "rules": {
+    "@next/next/no-document-import-in-page": "off",
     "@typescript-eslint/explicit-function-return-type": 0,
     "@typescript-eslint/explicit-member-accessibility": 0,
     "@typescript-eslint/indent": 0,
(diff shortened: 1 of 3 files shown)


k8hansen (approved):
Great job!!

Wonderful description in the PR, thank you for taking the time to make this very clear.
🚢

PR #739, opened 2023-12-05
feat: 736 Feature Flag

k8hansen:


src/pages/_app.tsx
```diff
@@ -34,7 +36,7 @@ const CustomApp: FC<AppProps> = ({ Component, pageProps }) => {
   // }, [router.events]);
 
   return (
-    <>
+    <SSRProvider>
```
k8hansen, line 39: Do we still need this?

src/hooks/useFeatureToggle.ts
```diff
@@ -0,0 +1,13 @@
...
+
+import { FeatureToggleContext } from '../context/featureToggleContext';
+
+const useFeatureToggle = () => {
+  const { enabledFeatures } = useContext(FeatureToggleContext);
+  const isEnabled = (featureName: string) =>
+    enabledFeatures.includes(featureName);
+
+  return [isEnabled];
+};
```
mewdriller, line 11: 💭 Maybe a shift in the shape of this to something like:

```tsx
import { useContext } from 'react';

import { FeatureFlagsContext } from '../context/featureFlagsContext';

const useFeatureFlags = (...ids: string[]) => {
  const { enabledFlags } = useContext(FeatureFlagsContext);

  return ids.map((id) => enabledFlags.includes(id));
};

export default useFeatureFlags;
```

tsconfig.json
```diff
@@ -22,5 +22,5 @@
     "types": ["@emotion/react/types/css-prop"]
   },
   "exclude": ["node_modules", ".next", "out"],
-  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"]
+  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", "src/flags.js"]
```
mewdriller, line 25: ⚠️ Should this be written in TypeScript? And/or it looks like this doesn't exist at this point?

src/hooks/useFeatureToggle.ts
```diff
@@ -0,0 +1,13 @@
+import { useContext } from 'react';
+
+import { FeatureToggleContext } from '../context/featureToggleContext';
+
+const useFeatureToggle = () => {
```
mewdriller, line 5: 💭 I would expect a name like `useFeatureFlags` for this probably.

src/context/FeatureToggleProvider.tsx
```diff
@@ -0,0 +1,21 @@
+import { FC } from 'react';
+
+import { FeatureToggleContext } from '~/context/featureToggleContext';
+
+type FeatureToggleProps = {
+  children: React.ReactNode;
+  enabledFeatures: string[];
+};
+
+const FeatureToggle: FC<FeatureToggleProps> = ({
```
mewdriller, line 10: 💭 I would expect a name like `FeatureFlagsProvider` for this.

src/context/featureToggleContext.ts
```diff
@@ -0,0 +1,5 @@
+import { createContext } from 'react';
+
+export const FeatureToggleContext = createContext({
```
mewdriller, line 3: 💭 I would expect a name like `FeatureFlagsContext` here.

src/pages/_app.tsx
mewdriller: It's needed for things within the version of `@react-aria/*`  we're using to properly render in SSR. In more recent versions of React there is a `useId()` hook that removes the need for it.

mewdriller (commented):
Here's an (outside) perspective on things. Let me know if you want to chat about the feedback at all.

src/pages/_app.tsx
k8hansen: Thanks, we JUST updated to the new version of React but we were getting errors when Sean had another solution to our flab problem so we added it back in. 

Looks like we can remove it when we update `@react-aria` then? cc: @nelsW70

tsconfig.json
k8hansen: Yes, thank you. There was an alleged reason this was just a JS file, but that was like 5 permutations ago, this is a leftover.

mewdriller (approved):
LGTM!

PR #753, opened 2023-12-08
feat(ui): Dynamic Next Image (B2-694)

mlam0527:
**Description**
- utilized image component from next/image to be responsive to screen size changes

Note:
The image uses "responsive" layout prop which allows the image to scale down based on width of the window. The image height and width has to be provided in order for this feature to work correctly.


nelsW70 (approved):
Nice work!

Issue #722, opened 2023-11-08
Organize Components


mlam0527:
As a developer, I want to navigate the components folder with ease by organizing components to their respective pages.

**Acceptance Criteria**:
- Create a common folder for common components
- Organize layout/header/footer components into separate folder
   - This is similar to how the Approach Page, Battery Page etc. are laid out
- HomePage Folder
   - place pngs into new image folder (to follow current convention)

nothingalike:
link to pages: https://drive.google.com/drive/u/1/folders/1aPzIHF3hVpFt33slmPE1EoxUYD5At4uV

PR #763, opened 2023-12-12
B2 722 - Organized the folder structure of components

JaviAir:
[closes](https://github.com/b2io/base2.io/issues/722)

Before
image

After
image


sludemann:
Tested and reviewed the basic changes. No major issues, there's more that we could / would here but we'll tackle that in the re-architecture.

sludemann (approved):

PR #767, opened 2023-12-19
Chore: Framer Dependency Update

k8hansen:
The animation is a little different than in current prod, but in the interest of time we'd like to merge in the dependancy updates so we can work on other animations on the site.

## To Test

- pull branch
- `npm i`
- run dev
- build and run prod

Note animations on homepage hero, and page transitions.


nelsW70 (approved):
Sweet, nice work!

PR #769, opened 2023-12-19
761 Update Next version

nelsW70:
- update Next to latest version and the majority of Next Image usages


framer.d.ts
```diff
@@ -0,0 +1,7 @@
+import { ReactNode } from 'react';
+
+declare module 'framer-motion' {
```
k8hansen, line 3: 💭 do we still need this after framer updates

k8hansen (approved):
LGTM! :shipit:

framer.d.ts
nelsW70: Nope, we do not

PR #773, opened 2023-12-20
771: fix double animation load on home screen

nelsW70:
removed motion.div from app.tsx as animation values are being passed in the Layout component and it appears to resolve the issue

**Test**
- run locally and verify that the home page does not double load

resolves [771](https://github.com/orgs/b2io/projects/15/views/1?pane=issue&itemId=48110836)


k8hansen (approved):
🤦‍♀️ Looks good, thanks!

PR #779, opened 2024-01-03
B2 778 delete storybook

JaviAir:
Purged all storybook related files, packages, text.

[closes](https://github.com/b2io/base2.io/issues/778)


nelsW70 (approved):
Nice! 

The only change was removing the Storybook reference from the readme so I pushed that and we're good.

PR #782, opened 2024-01-03
fix: One case study unseen error

JaviAir:
[closes](https://github.com/b2io/base2.io/issues/781)

### DIFF
--- a/src/components/CaseStudy/CaseStudyBottomNav.tsx
+++ b/src/components/CaseStudy/CaseStudyBottomNav.tsx
@@ -18,11 +18,17 @@ const getUnseenStudies = (
 };
 
 const pickTwoStudies = (
-  studies: CaseStudyPageConfig[],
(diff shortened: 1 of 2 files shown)


src/components/CaseStudy/CaseStudyBottomNav.tsx
```diff
@@ -18,11 +18,17 @@ const getUnseenStudies = (
 };
 
 const pickTwoStudies = (
-  studies: CaseStudyPageConfig[],
-  excludeId?: number,
+  unseenStudies: CaseStudyPageConfig[],
+  otherStudies: CaseStudyPageConfig[],
```
sludemann, line 22: Good work here, this definitely fixes the issue. I feel like the code is getting a little hard to follow so I'm going to push some quick changes to generalize the logic a bit.

sludemann (approved):

PR #802, opened 2024-01-08
B2 800 dev dependencies update - prettier/stylelint

mlam0527:
**Description**
- updated prettier and stylelint to current major versions


nelsW70 (approved):
Nice work!

PR #823, opened 2024-01-23
822: CardGrid accessibility issue

nelsW70:
- passed props to BasicCard in order to give children access to "as" prop and properly render as "li" elements

Before
Screenshot 2024-01-23 at 9 50 41 AM

After
Screenshot 2024-01-23 at 1 38 54 PM


mlam0527 (approved):
Good catch, LGTM 🚀

PR #825, opened 2024-01-23
824: ResultCard accessibility issue

nelsW70:
passed props to ResultCard in order to give children access to "as" prop and properly render as "li" elements

Before
Screenshot 2024-01-23 at 3 54 45 PM

After
Screenshot 2024-01-23 at 3 58 07 PM


mlam0527 (approved):
LGTM 🚀

PR #827, opened 2024-01-25
826:  fix ResultCard typography 

nelsW70:
- refactored ResultCard to reduce inline typography styling and prevent breakage

**Note**
The goal with this is not to make the styling match the designs but to prevent broken layouts and simplify the approach to typography until it can be addressed 

Resolves [826](https://github.com/orgs/b2io/projects/15/views/1?pane=issue&itemId=51183468)


mlam0527 (approved):
good work!

PR #842, opened 2024-01-29
Referrals info

nelsW70:


src/components/ServicesPage/Referrals.tsx
```diff
@@ -0,0 +1,28 @@
...
+          margin: ${spacing.xxl} 0 ${spacing.md};
+        `}
+      >
+        Referrals
+      </Heading>
+      <Heading as="h4" variant="h3">
+        Base Two is a small group of thoughtful developers with no dedicated
+        sales team. We generate work through word of mouth and satisfied
+        clients. Know of someone who could use help with a project? We do offer
+        referral fees in certain situations. Drop us a line and we&#39;ll be
```
tborres, line 23: I'm not sure of the word `fees` here, which has a bit of a negative connotation. Maybe referral "benefits", "perks", "rewards". I'm not sure what works best.

tborres (commented):
Great first pass, and I saw the feedback in Slack.

I think that this may have a better home on the Contact page, which is where somebody without a clear project may be more likely to go.

Here's another verbiage option to consider:

>At Base Two, we are a team of passionate software engineers who love what we do - as such we don’t have a sales team. Our sales primarily come from satisfied customers and word of mouth. Do you know someone who needs help with a project? We appreciate your referrals and we are happy to reward you for them in certain situations. Contact us and we’ll discuss the details.

src/components/ServicesPage/Referrals.tsx
nelsW70: Yeah, your verbiage is better and agreed on Contact form being a better place for it to live. 

updated

PR #861, opened 2024-02-09
860: Common section container

nelsW70:
- Created and implemented shared section container as identical styling was being applied in numerous instances across the site

Resolves [860](https://github.com/orgs/b2io/projects/15/views/1?pane=issue&itemId=52691291)


k8hansen (approved):
Great catch!

PR #889, opened 2024-06-04
Fix-871: Make CTA Design Clearer

k8hansen:


sludemann (commented):
Clicked through the previous, and reviewed the code. Overall I'm really liking the changes, I think the much stronger CTA approach does what it's supposed to do in a solid way.  I saw one weird bit with the animation where if you navigate and then go back, the on screen CTA's won't appear until you scroll a bit, but I don't see that as a blocker at all.  Overall good work to tie things together!

laurapadilla (commented):
I think the changes are great. The only 2 things I'd consider are:
1. Consider adjusting the CTA padding to `1rem 1.5rem` so that they look more like buttons. 
2. Consider adding a little arrow on the CTAs: `View More →`
3. Consider moving the Request Quote CTA, right now it feels a little out of place in the middle of the footer. Maybe even adding it as a link on the header as well might work?
Screenshot 2024-06-05 at 2 10 49 PM

laurapadilla (approved):
🦋 LGTM!

PR #893, opened 2024-06-07
#772: Console log warning from react-scroll

### LINKED ISSUE #772: console warning from react-scroll usage
Clicking react-scroll link on Services page triggers console warning "Listener must be a named function" warning

laurapadilla:
Closes #772 

I think the `Service` tag needs to be replaced with the Element tag from react-scroll. Since the warning only shows up in production, I'm not totally sure how to test this.

### DIFF
--- a/src/components/ServicesPage/OurServices.tsx
+++ b/src/components/ServicesPage/OurServices.tsx
@@ -1,7 +1,7 @@
 import { css } from '@emotion/react';
 import styled from '@emotion/styled';
 import type { FC } from 'react';
-import { Link } from 'react-scroll';
+import { Element, Link } from 'react-scroll';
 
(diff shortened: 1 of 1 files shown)


src/components/ServicesPage/OurServices.tsx
```diff
@@ -53,12 +53,12 @@ const ServicesContainer = styled.section`
   margin-top: ${spacing.xxl2};
 `;
 
-const Service = styled.div`
-  display: flex;
-  flex: 1 1 auto;
-  flex-wrap: wrap;
-  gap: ${spacing.xs};
-`;
+// const Service = styled.div`
```
k8hansen, line 56: Remove

k8hansen (approved):
Let's try it, we can always roll it back!

